PR #9, opened 2022-07-06
add support for some cudax libraries

ryanleary:
This adds basic support for curand, cufft, cusolver, cusparse, nvcc headers, and nvtx

### DIFF
--- a/cuda/runtime/BUILD.local_cuda
+++ b/cuda/runtime/BUILD.local_cuda
@@ -324,10 +324,23 @@ cc_library(
     ]),
 )
 
-# nvcc
+# nvptxcompiler
+cc_import(
+    name = "nvptxcompiler_so",
+    static_library = "cuda/lib64/libnvptxcompiler_static.a",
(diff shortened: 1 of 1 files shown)


cuda/runtime/BUILD.local_cuda
```diff
@@ -298,3 +298,158 @@ cc_library(
...
+    ] + if_linux([
+        ":curand_so",
+    ]) + if_windows([
+        ":curand_lib",
+    ]),
+)
+
+# nvcc
+cc_library(
+    name = "nvcc",
```
cloudhan, line 329: Why is this named as `nvcc`. I think nvptx will be more reasonable. It also needs nvptxcompiler_static.lib or libnvptxcompiler_static.a to work properly

ryanleary: good catch, thanks. updated.

cloudhan:
LGTM. Thanks for the contribution.

PR #11, opened 2022-07-11
add defensive wiring for header-only libraries

ryanleary:
if you specify a `cuda_library` with no sources, an error is thrown here due to there being nothing to link. This adds a check that `library_to_link` is not None before proceeding.


cloudhan:
Great thanks for the contribution. I once faced the similar issue and that is why you see assignments to ternary experssion. Wondering if this is version dependent behaviour of `linking_outputs.library_to_link`.

PR #21, opened 2022-10-11
Disable arch_native by default

jsharpe:
It is common to compile on machines that don't have a GPU installed e.g. RBE build nodes. Also setting default compilation to use the native arch is not as reproducible because it is dependent upon the hardware of the build node that the compiler is invoked on.

Turn off the arch_native feature by default to make the default configuration more reproducible.


cloudhan:
Thank. It is also problematic for clang toolchain, since it does not support any equivalent to `-arch=native`

PR #25, opened 2022-11-03
refactor!: make structure consistent w/ rule guidelines

ryanleary:
Initial work to align the repository with style guide outlined here: https://bazel.build/rules/deploying

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,36 +1,23 @@
-# CUDA Rules for [Bazel](https://bazel.build)
+# CUDA rules for [Bazel](https://bazel.build)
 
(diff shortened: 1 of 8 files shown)


jsharpe (approved):

ryanleary:
I believe this will be a breaking change for anyone already using this.. We'll want to try to get any other breaking changes out of the way asap

jsharpe:
We could add forwarding functions that print deprecation messages if we really wanted to - but given there have been no formal releases yet and semver says that any release 0.x.y may introduce breaking changes (https://semver.org/#spec-item-4) then I don't think we need to worry.

PR #30, opened 2022-11-03
Move if_linux and if_windows into rules_cuda

jsharpe:
There is no dependency on the toolchain for these macros so they can be placed in rules_cuda directly.


ryanleary (approved):
lgtm

PR #31, opened 2022-11-03
Default cuda to on

jsharpe:
The rules should ideally work with no bazel flags. Having an extra opt-in step by default will likely confuse new users as to why cuda isn't building if the config flag isn't set to enable the rules.

### DIFF
--- a/cuda/BUILD.bazel
+++ b/cuda/BUILD.bazel
@@ -23,7 +23,7 @@ toolchain_type(name = "toolchain_type")
 # Set with --@rules_cuda//cuda:enable
 bool_flag(
     name = "enable",
-    build_setting_default = False,
+    build_setting_default = True,
 )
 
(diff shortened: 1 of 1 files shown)


ryanleary (commented):
I would suggest adding a sentence in the `README.md` to indicate that this is the default behavior in the `### Available Flags` section

cloudhan:
I think we might also better to update the `if_cuda` example to teach users how to add their flag to control the enabling of cuda in their specific code. Otherwise, ppl tend to rely on existing thing and it will results a monolithic control like that in tensorflow.

Analogs to CMake, our flag control `CUDA_FOUND=ON`, their flag control `MY_PROJECT_USE_CUDA=ON`

ryanleary:
Closing as this was incorporated into #54.

PR #33, opened 2022-11-03
Build fix

jsharpe:
Apologies previous PR didn't quite build correctly. This fixes that mistake.


ryanleary:
I should've caught it. Thanks for the quick fix.

PR #36, opened 2022-11-03
Initial bzlmod support

jsharpe:
Adds bzlmod support to rules_cuda.


jsharpe:
This isn't working at the moment - it fails to find the registered toolchain.

fmeum:
@jsharpe I think that you are hitting https://github.com/bazelbuild/bazel/issues/16607. If I comment out the `platforms` `http_archive` in `repositories.bzl`, the toolchain is found.

Before:
```starlark
$ bazel query @local_cuda//toolchain/clang:clang-local-toolchain --output=build --enable_bzlmod
toolchain(
  name = "clang-local-toolchain",
  visibility = ["//visibility:public"],
  exec_compatible_with = ["@~ext~platforms//cpu:x86_64"],
  toolchain_type = "//cuda:toolchain_type",
  target_compatible_with = ["@~ext~platforms//cpu:x86_64"],
  target_settings = ["//cuda:is_enabled", "//cuda:compiler_is_clang"],
  toolchain = "@~ext~local_cuda//toolchain/clang:clang-local",
)
```

After:
$ bazel query @local_cuda//toolchain/clang:clang-local-toolchain --output=build --enable_bzlmod
toolchain(
  name = "clang-local-toolchain",
  visibility = ["//visibility:public"],
  exec_compatible_with = ["@platforms~0.0.4//cpu:x86_64"],
  toolchain_type = "//cuda:toolchain_type",
  target_compatible_with = ["@platforms~0.0.4//cpu:x86_64"],
  target_settings = ["//cuda:is_enabled", "//cuda:compiler_is_clang"],
  toolchain = "@~ext~local_cuda//toolchain/clang:clang-local",
)
```

**Comment by Wyverald** [comment]
The problem is here:
https://github.com/bazel-contrib/rules_cuda/blob/8a515347af61a83187c4dbf700aa1f96c0575639/cuda/private/repositories.bzl#L176

This generates a BUILD file inside `@@~ext~local_cuda`, using a template that contains these lines:
https://github.com/bazel-contrib/rules_cuda/blob/8a515347af61a83187c4dbf700aa1f96c0575639/cuda/templates/BUILD.local_toolchain_clang#L34-L39

While it's true that nobody's `use_repo`ing `@@~ext~platforms`, this repo is still visible to all the repos generated by the same extension under its declared name (`platforms`). This name binding intentionally has higher priority than `@@platforms~0.0.4` (see logic at https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/skyframe/RepositoryMappingFunction.java;l=198-204;drc=bdcd65d878dc7aff006a5c6d86268d76b536c9a5)

The fix should be simple -- instead of calling `rules_cuda_dependencies()` in your module extension, just call `local_cuda(name = "local_cuda")`.

**Comment by meteorcloudy** [comment]
@Wyverald I don't think we need to fix anything from Bazel side, but maybe we can give some warnings if any well-known repos are generated by module extensions? Especially, `platforms`, and other repos containing platform and toolchain definitions, otherwise, there could be easily mysterious mismatching errors.

**Comment by cloudhan** [comment]
> I don't think we need to fix anything from Bazel side

I think the spooky action at a distance must be eliminated.

**Comment by Wyverald** [comment]
It's unclear how this could be "fixed", which is why we defined the behavior to be "prioritize the generated repo over the bazel_dep of the hosting module". Any ideas would be welcome.

File: cuda/extensions.bzl [hunk 1]
```diff
@@ -0,0 +1,8 @@
+"""Entry point for extensions used by bzlmod."""
+
+load("//cuda:repositories.bzl", "rules_cuda_dependencies")
+
+def _init(module_ctx):
+    rules_cuda_dependencies()
```
meteorcloudy, line 6: Blindly loading all your transitive dependencies in a module extension is a terrible idea. Because all of them will be "internal dependencies" of rules_cuda, and may introduce duplicated dependencies in the whole dependency graph.
@cloudhan @jsharpe This is what should be eliminated.

meteorcloudy (changes requested):

cuda/extensions.bzl
jsharpe: @meteorcloudy my mistake is so obvious now, but after having taken a short break from writing bzlmod code I'd misremembered the migration path of being able to add code in `WORKSPACE.bzlmod` vs putting it in the init function of a module. 

I suspect this might be a common mistake during migration. So any warnings that can be pulled out would be very useful, or if not, a 'lint' mode that searches for these conflicts would help.

jsharpe:
TODO: need to add a CI job to build with bzlmod enabled

jsharpe:
@ryanleary one thing this will change about the release process is a need to bump the version number in MODULE.bazel as part of the release process. I've not looked at the release scripts in detail to know how to do this automatically. Would this be easy to add?

MODULE.bazel
```diff
@@ -0,0 +1,18 @@
+module(
+    name = "rules_cuda",
+    compatibility_level = 1,
+    version = "0.1.1",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.2.1")
+bazel_dep(name = "platforms", version = "0.0.4")
+bazel_dep(name = "rules_cc", version = "0.0.2")
```
cloudhan, line 9: Is rules_cc actively maintained? https://github.com/bazelbuild/rules_cc/issues/86

jsharpe: AFAIK adding a bazel_dep on `rules_cc` is how you express to bzlmod the fact that you are using cc rules in your module.  e.g. https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/abseil-cpp/20211102.0/MODULE.bazel

The convention seems to be that all modules using cc rules are adding a dep on rules_cc as a module.

cloudhan: This is weird, because we don't actually load anything from `@rules_cc`. Can we leave some comment about the weirdness here.

@meteorcloudy @Wyverald Would you mind to @ someone that is familiar with `rules_cc` to make it clear of the current state of it and correct practice of depending on bazel builtin rules?

Wyverald: hmm? if you don't use rules_cc, you don't need to depend on it.

It's also in a bit of a weird state right now since we modularized it, but most of the actual logic is still shipped with the Bazel binary, so the dependency is a bit ... cosmetic. @oquenchil can answer question in detail.

jsharpe: I guess the question is one of forward compatibility - if the plan is to remove the builtin `cc_*` targets and require them being loaded from rules_cc in the future then it is more future compatible to be loading them now even though they just forward through to the builtin rules as it won't break in a future version of bazel (and thus break the BCR entry for a newer bazel)

fmeum: @jsharpe It would break anyway as the pinned version of `rules_cc` still just delegates to the native rules. Supporting a future version of Bazel without these native rules would at least require a version update. At that point maintaining forward compatibility becomes somewhat moot.

jsharpe: Yes but it could be fixed fairly trivially by a bazel_dep in the root module to bump the version of `rules_cc` to one that fixes it for the version of bazel that removes the native rule?

cloudhan (approved):
Thanks for the contribution!

PR #38, opened 2022-11-03
Add thrust and cub repos

jsharpe:
As thrust can also be used via an OpenMP backend, make this available as repository outside of the toolchain.

### DIFF
--- a/cuda/private/repositories.bzl
+++ b/cuda/private/repositories.bzl
@@ -210,4 +210,22 @@ def rules_cuda_dependencies():
         ],
     )
 
+    maybe(
+        name = "thrust",
+        repo_rule = http_archive,
+        build_file = "@rules_cuda//third_party:thrust.BUILD",
+        sha256 = "d021e37f5aac30fd1b9737865399feb57db8e601ae2fc0af3cd41784435e9523",
+        strip_prefix = "thrust-1.17.2",
+        urls = ["https://github.com/NVIDIA/thrust/archive/refs/tags/1.17.2.tar.gz"],
+    )
+
+    maybe(
+        name = "cub",
+        repo_rule = http_archive,
+        build_file = "@rules_cuda//third_party:cub.BUILD",
+        sha256 = "1013a595794548c359f22c07e1f8c620b97e3a577f7e8496d9407f74566a3e2a",
+        strip_prefix = "cub-1.17.2",
+        urls = ["https://github.com/NVIDIA/cub/archive/refs/tags/1.17.2.tar.gz"],
+    )
+
     _local_cuda(name = "local_cuda")
--- a/third_party/cub.BUILD
+++ b/third_party/cub.BUILD
@@ -1,15 +1,15 @@
 load("@rules_cc//cc:defs.bzl", "cc_library")
-load("@rules_license//rules:license.bzl", "license")
+#load("@rules_license//rules:license.bzl", "license")
 
-package(default_applicable_licenses = [":license"])
+#package(default_applicable_licenses = [":license"])
 
-license(
-    name = "license",
-    license_kinds = [
-        "@rules_license//licenses/spdx:BSD-3-Clause",
-    ],
-    license_text = "LICENSE.TXT",
-)
+#license(
+#    name = "license",
+#    license_kinds = [
+#        "@rules_license//licenses/spdx:BSD-3-Clause",
+#    ],
+#    license_text = "LICENSE.TXT",
+#)
 
 filegroup(
     name = "include-src",
--- a/third_party/cub.BUILD
+++ b/third_party/cub.BUILD
@@ -1,0 +1,45 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@rules_license//rules:license.bzl", "license")
+
+package(default_applicable_licenses = [":license"])
+
+license(
+    name = "license",
+    license_kinds = [
+        "@rules_license//licenses/spdx:BSD-3-Clause",
+    ],
+    license_text = "LICENSE.TXT",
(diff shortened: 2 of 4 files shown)


jsharpe:
I may have missed some includes in thrust - the set here are just from what I've used on another project and so may be missing some for some functionality.
Have also got rules_license rules commented out here; its probably not quite stable enough yet to take a dependency on rules_license but when we do we'll want to add those back in.

cloudhan:
The problem here is the library is freezed for a single version and cannot be upgraded/downgraded by the user easily. Might better to stick with cuda toolkit bundled version. There are so many libraries and versions there, this should belong to a seperate repo that provides CUDA related libraries recipes, say cuDNN, TensorRT, etc.

jsharpe:
I don't necessarily think it should be a separate repo but could be a separate workspace contained within this repo. Ultimately it'll become a bzlmod registry for all these libraries.

We can add version selection to this quite easily as an argument to the dependencies macro. The motivation for adding this is two-fold; to allow use of  the downloadable tarballs for the cuda toolchain and to allow use of thrust / cub in the case that the user has disabled the cuda toolchain. Its perfectly valid to use thrust with an openmp backend and so this allows the user to do this without needing a cuda install to actually be present.

cuda/private/repositories.bzl
```diff
@@ -210,4 +210,22 @@ def rules_cuda_dependencies():
...
+
+    maybe(
+        name = "cub",
+        repo_rule = http_archive,
+        build_file = "@rules_cuda//third_party:cub.BUILD",
+        sha256 = "1013a595794548c359f22c07e1f8c620b97e3a577f7e8496d9407f74566a3e2a",
+        strip_prefix = "cub-1.17.2",
+        urls = ["https://github.com/NVIDIA/cub/archive/refs/tags/1.17.2.tar.gz"],
+    )
+
```
cloudhan, line 230: They are not mandatory, please move them into `third_party/` also. Otherwise, our core rules depends on not in use remote files. Lets keep them minimal.

third_party/cub.BUILD
```diff
@@ -0,0 +1,45 @@
...
+        "cub/block/*.cuh",
+        "cub/block/specializations/*.cuh",
+        "cub/device/*.cuh",
+        "cub/device/dispatch/*.cuh",
+        "cub/grid/*.cuh",
+        "cub/warp/*.cuh",
+        "cub/warp/specializations/*.cuh",
+        "cub/iterator/*.cuh",
+        "cub/thread/*.cuh",
+        "cub/detail/*.cuh",
```
cloudhan, line 37: Any reason to make the glob so detailed? This prone to be affected by upstream implemtation details change. Why not just `glob([cub/**/*.cuh])`.

third_party/thrust.BUILD
```diff
@@ -0,0 +1,37 @@
...
+#        "@rules_license//licenses/spdx:Apache-2.0",
+#        "@rules_license//licenses/spdx:BSL-1.0",
+#    ],
+#    license_text = "LICENSE",
+#)
+
+filegroup(
+    name = "include-src",
+    srcs = glob([
+        "thrust/*.h",
```
cloudhan, line 18: This line is redundent.

third_party/cub.BUILD
```diff
@@ -0,0 +1,45 @@
...
+#license(
+#    name = "license",
+#    license_kinds = [
+#        "@rules_license//licenses/spdx:BSD-3-Clause",
+#    ],
+#    license_text = "LICENSE.TXT",
+#)
+
+filegroup(
+    name = "include-src",
```
cloudhan, line 15: Maybe call them `include-hdrs`.

third_party/thrust.BUILD
```diff
@@ -0,0 +1,37 @@
...
+#    name = "license",
+#    license_kinds = [
+#        "@rules_license//licenses/spdx:Apache-2.0",
+#        "@rules_license//licenses/spdx:BSL-1.0",
+#    ],
+#    license_text = "LICENSE",
+#)
+
+filegroup(
+    name = "include-src",
```
cloudhan, line 16: Same

ryanleary:
Wanted to point out that `cub`, `thrust`, and `libcudacxx` have all been coalesced into https://github.com/NVIDIA/cccl (all three projects remain header-only) recently, so if this integration moves forward, it should use that repository.

Would it make more sense to add build rules to `cccl` directly? I can reach out to that team - though it may be somewhat onerous to reach full coverage of the various flags/preprocessor directives

cloudhan:
The current blocker is merely how do we avoid `<dir_of_local_cuda>/cuda/include/{cub,thrust}` been exposed to the `:cuda_headers`, https://github.com/bazel-contrib/rules_cuda/blob/7a29239f41bbe1662104eb1b4ccf7bcd846ce3de/cuda/runtime/BUILD.local_cuda#L26-L30 

Or more generally, we need a confugrable way to piece together different packages (like in `<dir_of_local_cub>` and `<dir_of_local_nvcc>`, etc) to form a unified cuda toolkit directory, and use it as the `<dir_of_local_cuda>/cuda`. This will be useful for #72 as well

jsharpe:
I think ultimately this will be replaced by bzlmod modules: https://github.com/bazelbuild/bazel-central-registry/pull/1058.
@ryanleary ultimately yes if the CCCL team can own releasing this to the BCR with build rules then that would be the best solution for the community.

cloudhan:
@jsharpe We still need to exclude the folder under <toolkit_root>/include/{cub,thrust} to make it clean. Otherwise the deps order might cause incorrect include, similar to #105

PR #44, opened 2022-11-04
Make examples its own workspace

### LINKED ISSUE #39: CI should test rules_cuda as an external workspace
Currently the examples are in rules_cuda and so Label resolution can be wrong in changes and these still build correctly but break an external consumer of the ruleset. We should probably migrate examples to its own WORKSPACE and consume rules_cuda as an external repository to test these in the same way a user would use the rules.

jsharpe:
Closes #39

### DIFF
--- a/examples/.bazelrc
+++ b/examples/.bazelrc
@@ -1,0 +1,15 @@
+# Convenient flag shortcuts.
+build --flag_alias=enable_cuda=@rules_cuda//cuda:enable
+build --flag_alias=cuda_archs=@rules_cuda//cuda:archs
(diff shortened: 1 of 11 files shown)


examples/.bazelrc
```diff
@@ -0,0 +1,15 @@
+# Convenient flag shortcuts.
```
cloudhan, line 1: Let's symlink this file, from top level, instead of copying it. Just like that in docs directory.

ryanleary: Is the top-level `.bazelrc` even needed?

ryanleary:
I will merge and then add the symlink in a subsequent MR. Adding the release automation is blocked by this

PR #48, opened 2022-11-11
Add features for relaxed-constexpr and extended-lambda

jsharpe:


cloudhan:
How about make them as enabled by default. I don't think they will cause any problem. Any concerns?

jsharpe:
> How about make them as enabled by default. I don't think they will cause any problem. Any concerns?

Done, and also added to the msvc nvcc toolchain too.

cuda/private/toolchain_configs/nvcc.bzl
```diff
@@ -387,6 +387,34 @@ def _impl(ctx):
...
+    nvcc_extended_lambda_feature = feature(
+        name = "nvcc_extended_lambda",
+        enabled = True,
+        flag_sets = [
+            flag_set(
+                actions = [
+                    ACTION_NAMES.cuda_compile,
+                    ACTION_NAMES.device_link,
+                ],
+                flag_groups = [flag_group(flags = ["--extended-lambda"])],
```
cloudhan, line 413: Hmm, some old version of nvcc only supports [--expt-extended-lambda](https://docs.nvidia.com/cuda/archive/10.1/cuda-compiler-driver-nvcc/#options-for-altering-compiler-linker-behavior-expt-extended-lambda)

so you might use `["--extended-lambda" if nvcc_version_ge(ctx, some, version) else "--expt-extended-lambda"]`

cloudhan: `nvcc_version_ge(ctx, 10, 2)`

https://docs.nvidia.com/cuda/archive/10.2/cuda-compiler-driver-nvcc/#options-for-altering-compiler-linker-behavior-expt-extended-lambda

PR #49, opened 2022-11-12
Setup CI pipeline for testing docs generation

### LINKED ISSUE #47: Test docs generation in CI pipelines
We don't generate the docs and host it somewhere at the moment. But it is better to test the docs generation will not be breaked by the change. We can also extend it to support docs publish later.

cloudhan:
Closes #47

### DIFF
--- a/.github/workflows/docs-tests.yaml
+++ b/.github/workflows/docs-tests.yaml
@@ -1,0 +1,30 @@
+name: Test Generating Docs
+
+on:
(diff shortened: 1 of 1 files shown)


cloudhan:
@jsharpe How do you think, this is an very loose check, so that we ensure the docs gen won't break on PR.

jsharpe (approved):
LGTM

PR #54, opened 2022-12-21
Enable rules_cuda by default, add docs

ryanleary:
Follows #31 with minor changes to the `if_cuda` and expansion of related documentation.


cloudhan (approved):
Thanks.

PR #59, opened 2023-01-03
Add option to add files to the toolchain configuration

jsharpe:
Adds the ability to specify files that are needed by the compiler toolchain which is the first step towards being able to create a hermetic cuda toolchain.


cloudhan (approved):
LGTM. Thanks.

PR #60, opened 2023-01-03
Use merge_cc_infos to calculate common flags

jsharpe:
This fixes a bug where the transitive flags aren't propagated correctly in all cases. Notably adds the _builtin_deps dependency to pick up context from the configured `//cuda:runtime` dependency.

### DIFF
--- a/cuda/private/cuda_helper.bzl
+++ b/cuda/private/cuda_helper.bzl
@@ -157,6 +157,7 @@ def _create_common_info(
         host_compile_flags = [],
         host_link_flags = [],
         ptxas_flags = [],
+        transitive_cc_info = None,
         transitive_linking_contexts = []):
     """Constructor of the common object.
 
@@ -193,6 +194,7 @@ def _create_common_info(
         host_compile_flags = host_compile_flags,
         host_link_flags = host_link_flags,
         ptxas_flags = ptxas_flags,
+        transitive_cc_info = transitive_cc_info,
         transitive_linker_inputs = [ctx.linker_inputs for ctx in transitive_linking_contexts],
         transitive_linking_contexts = transitive_linking_contexts,
     )
@@ -204,18 +206,21 @@ def _create_common(ctx):
     """
     attr = ctx.attr
 
+    all_cc_deps = [dep for dep in attr.deps if CcInfo in dep]
+    if hasattr(attr, "_builtin_deps"):
+        all_cc_deps.extend([dep for dep in attr._builtin_deps if CcInfo in dep])
+
+    merged_cc_info = cc_common.merge_cc_infos(cc_infos = [dep[CcInfo] for dep in all_cc_deps])
+
     # gather include info
-    includes = []
+    includes = merged_cc_info.compilation_context.includes.to_list()
     system_includes = []
     quote_includes = []
     quote_includes.extend(_resolve_workspace_root_includes(ctx))
     for inc in attr.includes:
         system_includes.extend(_resolve_includes(ctx, inc))
-    for dep in attr.deps:
-        if CcInfo in dep:
-            includes.extend(dep[CcInfo].compilation_context.includes.to_list())
-            system_includes.extend(dep[CcInfo].compilation_context.system_includes.to_list())
-            quote_includes.extend(dep[CcInfo].compilation_context.quote_includes.to_list())
+    system_includes.extend(merged_cc_info.compilation_context.system_includes.to_list())
+    quote_includes.extend(merged_cc_info.compilation_context.quote_includes.to_list())
 
     # gather header info
     public_headers = []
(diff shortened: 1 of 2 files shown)


cuda/private/rules/cuda_library.bzl
```diff
@@ -81,6 +81,18 @@ def _cuda_library_impl(ctx):
...
+        CcInfo(
+            compilation_context = compilation_ctx,
+            linking_context = linking_ctx,
+        ),
+    ]
+
+    if hasattr(ctx.attr, "_builtin_deps"):
+        cc_infos.extend([dep[CcInfo] for dep in ctx.attr._builtin_deps if CcInfo in dep])
+
+    cc_info = cc_common.merge_cc_infos(cc_infos = cc_infos, direct_cc_infos = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep])
```
cloudhan, line 94: I am not sure if this is correct here. The `direct_cc_infos`, as per the docs,

> List of CcInfos to be merged, whose headers will be exported by the direct fields in the returned provider.

I think cc info from deps is not direct.

cuda/private/cuda_helper.bzl
```diff
@@ -204,18 +204,21 @@ def _create_common(ctx):
     """
     attr = ctx.attr
 
+    all_ccdeps = [dep for dep in attr.deps if CcInfo in dep]
+    if hasattr(attr, "_builtin_deps"):
+        all_ccdeps.extend([dep for dep in attr._builtin_deps if CcInfo in dep])
+
+    cc_info = cc_common.merge_cc_infos(cc_infos = [dep[CcInfo] for dep in all_ccdeps])
```
cloudhan, line 211: nit: I think `all_ccdeps` and `cc_info` better to be renamed to `all_cc_deps` and `merged_cc_info` respectively.

cuda/private/rules/cuda_library.bzl
cloudhan, line 94: Assume you return `merged_cc_info` from the `cuda_common.merged_cc_info`, I think this should be written/shorten as
```starlark
cc_info = cc_common.merge_cc_infos(cc_infos = common.merged_cc_info, direct_cc_infos = [CcInfo(
            compilation_context = compilation_ctx,
            linking_context = linking_ctx,
        )])
```

cloudhan: Take https://bazel.build/rules/lib/CompilationContext#direct_headers as an example:
> Returns the list of modular headers that are declared by this target. This includes both public headers (such as those listed in "hdrs") and private headers (such as those listed in "srcs").

jsharpe: Yep, agreed; I will revisit this.

cloudhan: Maybe call it as `cuda_common.merged_transitive_cc_info` to make it clear that it is not direct :)

ryanleary:
possible to have a test to expose this issue?

cuda/private/rules/cuda_library.bzl
```diff
@@ -90,8 +92,8 @@ def _cuda_library_impl(ctx):
             pic_objects = pic_objects,
         ),
         CcInfo(
-            compilation_context = compilation_ctx,
-            linking_context = linking_ctx,
+            compilation_context = cc_info.compilation_context,
+            linking_context = cc_info.linking_context,
         ),
```
jsharpe, line 97: IIRC simply returning cc_info here didn't seem to work but not sure why?

jsharpe:
> possible to have a test to expose this issue?

Not yet! It has shown up in my hacky attempts to get a hermetic version of the cuda toolchain working.
The main issue was that `_builtin_deps` wasn't added as a transitive dependency and so any flags from there weren't made available to downstream targets.

cuda/private/rules/cuda_library.bzl
jsharpe: I went with `transitive_cc_info` to mirror `transitive_linking_contexts` which already existed.

cuda/private/cuda_helper.bzl
jsharpe: Done

cuda/private/rules/cuda_library.bzl
cloudhan: Maybe a bug in bazel itself, let ignore it ;)

cloudhan (approved):
Great simplification! Thanks for contribution.

cloudhan:
@jsharpe Do you have any concern? Can we merge it?

jsharpe:
> @jsharpe Do you have any concern? Can we merge it?

No concerns; this is ready to merge.

PR #64, opened 2023-01-06
Add publish-to-bcr configuration

jsharpe:
Adds configuration for the publish-to-bcr app which auto publishes a PR to the bazel-central-registry whenever a release is made.

### DIFF
--- a/.bcr/config.yml
+++ b/.bcr/config.yml
@@ -1,0 +1,3 @@
+fixedReleaser:
+  login: jsharpe
+  email: james.sharpe@zenotech.com
--- a/.bcr/config.yml
+++ b/.bcr/config.yml
@@ -1,3 +1,0 @@
-fixedReleaser:
-  login: jsharpe
-  email: james.sharpe@zenotech.com
--- a/.bcr/metadata.template.json
+++ b/.bcr/metadata.template.json
@@ -1,0 +1,13 @@
+{
+  "homepage": "https://github.com/bazel-contrib/rules_cuda",
+  "maintainers": [
+    {
+      "email": "james.sharpe@zenotech.com",
+      "github": "jsharpe",
+      "name": "James Sharpe"
+    }
+  ],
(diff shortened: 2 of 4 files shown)


.bcr/config.yml
```diff
@@ -0,0 +1,3 @@
+fixedReleaser:
+  login: jsharpe
+  email: james.sharpe@zenotech.com
```
jsharpe, line 3: We may be able to remove this if bazel-contrib/bazel-central-registry gets the publish-to-bcr app configured on it.

.bcr/presubmit.yml
```diff
@@ -0,0 +1,12 @@
+# We recommend included a bcr test workspace that exercises your ruleset with bzlmod.
+# For an example, see https://github.com/aspect-build/bazel-lib/tree/main/e2e/bzlmod.
+#bcr_test_module:
```
jsharpe, line 3: We can't run tests in the registry until we have a hermetic toolchain as there will be no cuda install in bazelbuild CI

.bcr/config.yml
jsharpe: App should be configured for all repos in bazel-contrib so this should work ok without a fixed releaser.

.bcr/metadata.template.json
```diff
@@ -0,0 +1,13 @@
+{
+  "homepage": "https://github.com/bazel-contrib/rules_cuda",
+  "maintainers": [
+    {
+      "email": "james.sharpe@zenotech.com",
+      "github": "jsharpe",
+      "name": "James Sharpe"
+    }
```
jsharpe, line 8: @cloudhan @ryanleary note that this is the maintainer of the BCR entry not necessarily the repo; feel free to add yourself to this list if you want but I'm happy to be listed here.

jsharpe:
Once this is merged we can try cutting a release and it should post a PR to the central registry to add rules_cuda as a module.

cloudhan (approved):

PR #65, opened 2023-02-03
Add host compiler flag to device link step

jordancharest:
As is, nvlink will use the system compiler configuration to produce objects, resulting in an error when building //rules_cuda/examples/rdc:main  in 2 cases:
* When using a custom gcc toolchain that is a different version than your system install, nvcc will fail to find gcc after nvlink-ing
* When cross-compiling, nvlink will produce objects for the host architecture instead of the target architecture

This change is also likely needed in nvcc_msvc.bzl, but I have no way to test.

### DIFF
--- a/cuda/private/toolchain_configs/nvcc.bzl
+++ b/cuda/private/toolchain_configs/nvcc.bzl
@@ -68,7 +68,10 @@ def _impl(ctx):
         enabled = True,
         flag_sets = [
             flag_set(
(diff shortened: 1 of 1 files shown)


cloudhan:
Good catch. Lets ignore msvc then, becase brief skimming through docs, CUDA have not mention the cross compilation for windows, and nvcc on windows only supports msvc, there will no way to use a custom msvc toolchain.

PR #66, opened 2023-02-05
Allow multiple 'local' installs of cuda to be specified

jsharpe:
Exposes the `local_cuda` repository_rule to allow additional installs of cuda to be installed as toolchains.
e.g.
```
local_cuda(name="cuda_11_4", toolkit_path = "/usr/local/cuda-11.4")
```


cuda/extensions.bzl
```diff
@@ -2,7 +2,23 @@
 
 load("//cuda/private:repositories.bzl", "local_cuda")
 
+cuda_toolkit = tag_class(attrs = {
+    "name": attr.string(doc = "Name for the toolchain repository"),
```
fmeum, line 6: This brings up an interesting and not fully explored aspect of Bzlmod. I *think* that having a `name` attribute on a tag class that ends up as a repository name verbatim should be considered an antipattern and avoided if possible: If names collide between two transitive dependencies that both use rules_cuda, users can't really do anything about their broken build.

I described this issue and a possible workaround in https://github.com/bazelbuild/rules_go/pull/3443. In the particular situation of rules_cuda, a simpler solution could be to ignore tags of modules other than the root module (`module_ctx.is_root`). That simplifies the implementation for now and allows you to explore this design space in more detail in the future.

CC @Wyverald

fmeum (approved):

cuda/extensions.bzl
jsharpe: FYI this pattern came from @alexeagle 's PR to rules-template: https://github.com/bazel-contrib/rules-template/pull/46/files

cloudhan:
Is is still draft. I think this should good to go?

cuda/extensions.bzl
```diff
@@ -2,7 +2,28 @@
...
+    "toolkit_path": attr.string(doc = "Path to the CUDA SDK"),
+})
+
 def _init(module_ctx):
-    local_cuda(name = "local_cuda")
+    registrations = {}
+    for mod in module_ctx.modules:
+        for toolchain in mod.tags.local_toolchain:
+            if not mod.is_root:
+                fail("Only the root module may override the path for the local cuda toolchain")
```
jsharpe, line 15: Note that this is an explicit choice that a user has to define a cuda toolchain in their MODULE.bazel file. 
This is anticipating that we will be able to setup a fully hermetic toolchain and so the root MODULE will decide whether to use the locally installed cuda or the hermetic toolchain.

cloudhan:
It is weird that the CI failure cannot be reproduced on local machine with cuda 11.*

cloudhan:
@jsharpe Finally got sometime to figure this one out. The reason a mixture of CI and you impl. On CI we only installed `cuda-nvcc-10-1 cuda-cudart-dev-10-1`, then /usr/local/cuda is not configured so only /usr/local/cuda-10.1 exists. And in you change `cuda.local_toolchain(name = "local_cuda", toolkit_path = "/usr/local/cuda")` cause bazel creating a dangling reference. Can you change it default to `toolkit_path = ""`? So that the environment CUDA_PATH will be picked up.

examples/MODULE.bazel
```diff
@@ -4,9 +4,11 @@ module(
...
-bazel_dep(name = "rules_cuda", version = "0.1.1")
+bazel_dep(name = "rules_cuda", version = "0.1.3")
+local_path_override(module_name = "rules_cuda", path = "..")
 
-cuda_toolchain = use_extension("@rules_cuda//cuda:extensions.bzl", "local_toolchain")
-use_repo(cuda_toolchain, "local_cuda")
+cuda = use_extension("@rules_cuda//cuda:extensions.bzl", "toolchain")
 
-local_path_override(module_name = "rules_cuda", path = "..")
+cuda.local_toolchain(name = "local_cuda", toolkit_path = "/usr/local/cuda")
```
cloudhan, line 12: You need to make it default to `toolkit_path = ""` to let the CI pass.

cloudhan (approved):
Thanks.

PR #74, opened 2023-03-22
Various fixes to get rules_cuda to work with Lambda-stack

Lalaland:
The rules for this repository do not currently work with the following commonly used CUDA configuration: https://lambdalabs.com/lambda-stack-deep-learning-software

The reason is because the Lambda deep stack installs CUDA in /usr/bin, which breaks a couple of assumptions with this software.

In particular:
- /usr/lib64 is the wrong path here, /usr/lib/x86_64-linux-gnu/ is correct. 
- /usr/bin/** is invalid since it contains recursive symlinks

This fixes those two issues.

### DIFF
--- a/cuda/runtime/BUILD.local_cuda
+++ b/cuda/runtime/BUILD.local_cuda
@@ -11,7 +11,9 @@ filegroup(
         ":_cuda_header_files",
     ] + glob([
         "cuda/version.txt",
-        "cuda/bin/**",
+        "cuda/bin/nvcc",
+        "cuda/bin/nvlink",
+        "cuda/bin/bin2c",
         "cuda/lib64/**",
         "cuda/nvvm/**",
     ]),
--- a/cuda/runtime/BUILD.local_cuda
+++ b/cuda/runtime/BUILD.local_cuda
@@ -11,9 +11,7 @@ filegroup(
         ":_cuda_header_files",
     ] + glob([
         "cuda/version.txt",
-        "cuda/bin/nvcc",
-        "cuda/bin/nvlink",
-        "cuda/bin/bin2c",
+        "cuda/bin/**",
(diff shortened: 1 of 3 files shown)


cuda/runtime/BUILD.local_cuda
```diff
@@ -11,7 +11,9 @@ filegroup(
         ":_cuda_header_files",
     ] + glob([
         "cuda/version.txt",
-        "cuda/bin/**",
+        "cuda/bin/nvcc",
+        "cuda/bin/nvlink",
+        "cuda/bin/bin2c",
```
jsharpe, line 16: I suspect we also need ptxas here?

Lalaland: I just swapped out the design to bring back the **, which should fix this issue and make the code a bit less fragile.

cloudhan:
Just curious, in `BUILD.local_cuda`, all `cc_import` for `.so`s are non-versioned, will this work with Ubuntu's always versioned `.so.<version>`s?

Lalaland:
Ubuntu provides both versioned and unversioned copies so it works.

Lalaland:
@cloudhan Rebased and flipped. Note that you will have to manually trigger the workflows.

Lalaland:
@cloudhan Can you try rerunning it? I think you forgot to include the nvidia toolkit in your build script. I just added it so it should work now.

Lalaland:
Hmm. It looks like ubuntu ships an incompatible gcc and nvidia toolkit in their default packages? That's really quite lame. I added some code to the build script to try to downgrade the compiler.

Lalaland:
Sorry for the bother, but could someone try triggering the workflow again? I think I finally have the solution.

PR #81, opened 2023-03-28
Fix possibly conflict artifact names

### LINKED ISSUE #78: Attempted action contains artifacts not in previous action
Change `example/basic:kernel` to the following where `nested/kernel.cu` is a new empty file with duplicated basename

```starlark
cuda_library(
    name = "kernel",
    srcs = ["kernel.cu", "nested/kernel.cu"],
    hdrs = ["kernel.h"],
)
```

will trigger an error as follows:
```console
MandatoryInputs: Attempted action contains artifacts not in previous action (first 5): 
        basic/nested/kernel.cu
Previous action contains artifacts not in attempted action (first 5): 
        basic/kernel.cu
Outputs: are equal
ERROR: /home/cloud/rules_cuda/examples/basic/BUILD.bazel:5:13: for basic/_objs/kernel/kernel.o, previous action: action 'Compiling basic/kernel.cu', attempted action: action 'Compiling basic/nested/kernel.cu'
```

cloudhan:
Fixes #78

This changes to use basename indices to deduplicate as in cc rules.

### DIFF
--- a/cuda/private/actions/compile.bzl
+++ b/cuda/private/actions/compile.bzl
@@ -26,6 +26,9 @@ def compile(
 
     Returns:
         An compiled object `File`.
+
+    Notes:
+        The rules should call this action only once in case srcs have non-unique basenames.
     """
     actions = ctx.actions
     host_compiler = cc_toolchain.compiler_executable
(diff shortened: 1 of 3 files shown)


docs/build-docs.sh
```diff
@@ -21,7 +21,7 @@ function compose-versioned-site {
     python versioning.py generated/ --force
 
     printf "\nRun following command to update version list then serve locally:\n\n"
-    printf "\tpython -m http.server -d generate/\n\n"
+    printf "\tpython -m http.server -d generated/\n\n"
```
jsharpe, line 24: This doc fix should be a separate PR.

cuda/private/actions/compile.bzl
```diff
@@ -34,14 +37,27 @@ def compile(
...
+    ret = []
+    for src, basename, basename_index in src_and_indexed_basenames:
+        filename = None
         filename = cuda_helper.get_artifact_name(cuda_toolchain, artifact_category_name, basename)
+        # Objects are placed in _objs/<tgt_name>/<filename>.
+        # For files with the same basename, say srcs = ["kernel.cu", "foo/kernel.cu", "bar/kernel.cu"], we get
+        # _objs/<tgt_name>/0/kernel.<ext>, _objs/<tgt_name>/1/kernel.<ext>, _objs/<tgt_name>/2/kernel.<ext>.
+        # Otherwise, the index is not presented.
+        if basename_counter[basename] > 1:
+            filename = "{}/{}".format(basename_index, filename)
```
jsharpe, line 60: Could we instead go to `kernel.0.<ext>`, `kernel.1.<ext>` instead to avoid introducing the extra inode usage on the filesystem that creating extra directories incurs?

cloudhan: There will possible be
```
foo/kernel.cu -> kernel.0.cu ???
bar/kernel.cu -> kernel.1.cu
kernel.0.cu -> kernel.0.cu ???
```

Then the problem will be even annoying. 🤣

cloudhan: cc rules also has the same behavior.

jsharpe (approved):

PR #99, opened 2023-05-18
allow .cpp and .cc files in srcs

garymm:
NVCC is perfectly happy to compile these, and this is actually useful for my use case of having a single source that is compilable both as a regular cc_library and a cuda_library, depending on preprocessor directives inside the code.

### DIFF
--- a/cuda/private/rules/common.bzl
+++ b/cuda/private/rules/common.bzl
@@ -7,6 +7,7 @@ ALLOW_CUDA_HDRS = [
 ]
 
 ALLOW_CUDA_SRCS = [
+    ".cc",
+    ".cpp",
     ".cu",
-    ".cu.cc",
 ]


cloudhan:
I am not actually convinced with the idea of compiling `.cpp` with `cuda_library` (nvcc or clang) with cuda configuration.

The problem here is the pure cpp feature configuration in bazel builtin is rather complex. It contains a lot more features, like pgo, which are missing from `cuda_library`.

And the second issue is we don't want to build a full featured nvcc wrapper that do what nvcc can do, like compiling cpp files. Compiling cpp files with `cuda_library` is ambiguious. The `cuda_library` implies `-x cu` for both nvcc and clang. With `.cpp` files, do we want `nvcc -x cu file.cpp` or `nvcc -x cpp file.cpp`?

cloudhan:
@jsharpe Any idea?

I think if we want to allow the cpp files, then we want to do it with `nvcc -x cu file.cpp` with an additional rule attribute to gurad the behavior.

garymm:
Currently I'm working around this with a genrule to copy my `.cpp` files to `.cpp.cu`, which is just annoying, but it works.

garymm:
> The cuda_library implies -x cu for both nvcc and clang. With .cpp files, do we want nvcc -x cu file.cpp or nvcc -x cpp file.cpp?

I think `nvcc -x cu` makes sense. The use of `cuda_library` instead of `cc_library` clearly expresses the intent to use CUDA, regardless of the file extension.

garymm:
`-x cu` is already set here:
https://github.com/bazel-contrib/rules_cuda/blob/267a986d040ca27cc8c4553578eb2a04e892d32c/cuda/private/toolchain_configs/nvcc.bzl#L84

@cloudhan @jsharpe could you please reconsider this?

rnburn:
I really think this should be allowed.

Nvidia uses the .cpp extension in their own examples of separate compilation: https://developer.nvidia.com/blog/separate-compilation-linking-cuda-device-code/

It's going to be a hinderance to adoption if basic examples don't work.

cloudhan:
That is freaking annoying fact. And in ROCm world AMD folks are also doing this for their HIP sources. FYI, for users with CMake build system, the programmers need to set the source file language property manually, I am strongly opinionated and I think it is a bad practice.

But anyway. I think we shall relax the constraint.

cloudhan (approved):
Let's make everyone's life a little bit easier.

PR #106, opened 2023-06-02
Add NPP

lalten:
I have a .cuh file that does `#include <nppi_geometry_transforms.h>` and uses `nppiResize_8u_C1R_Ctx()`

Now I can just add `deps = ["@local_cuda//:nppig"],` to the target and it will link properly.

See https://docs.nvidia.com/cuda/npp/index.html

### DIFF
--- a/cuda/runtime/BUILD.local_cuda
+++ b/cuda/runtime/BUILD.local_cuda
@@ -516,7 +516,7 @@ _NPP_LIBS = {
 [
     cc_import(
         name = name + "_lib",
-        interface_library = "cuda/lib/x64/nppc.lib",
+        interface_library = "cuda/lib/x64/{}.lib".format(name),
         system_provided = 1,
         target_compatible_with = ["@platforms//os:windows"],
     )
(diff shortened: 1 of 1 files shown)


cuda/runtime/BUILD.local_cuda
```diff
@@ -473,3 +473,67 @@ cc_library(
...
+        shared_library = "cuda/lib64/lib{}.so".format(name),
+        target_compatible_with = ["@platforms//os:linux"],
+    )
+    for name in _NPP_LIBS.keys()
+]
+
+[
+    cc_import(
+        name = name + "_lib",
+        interface_library = "cuda/lib/x64/nppc.lib",
```
cloudhan, line 519: Should `nppc.lib` a mistake of `{}.lib` to be formatted with `name`?

lalten: Absolutely, thanks for the catch. I was not able to test this on Windows as you might have guessed ;P

PR #107, opened 2023-06-20
set nvcc flags for device debug info

garymm:
- set `nvcc --device-debug` when `compilation_mode=dbg`
- set `nvcc --generate-line-info` when `compilation_mode=fastbuild`

### DIFF
--- a/cuda/private/toolchain_configs/nvcc.bzl
+++ b/cuda/private/toolchain_configs/nvcc.bzl
@@ -329,7 +329,7 @@ def _impl(ctx):
         flag_sets = [
             flag_set(
                 actions = [ACTION_NAMES.cuda_compile],
-                flag_groups = [flag_group(flags = ["-Xcompiler", "-g1"])],
+                flag_groups = [flag_group(flags = ["--generate-line-info", "-Xcompiler", "-g1"])],
             ),
         ],
         provides = ["compilation_mode"],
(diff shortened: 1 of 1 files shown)


cloudhan:
Please don't associate it with dbg, instead, add a new feature `cuda-device-debug` and let the user to specify which target to enable the device debug symbol, or enable it globally on command line with `--feature=cuda-device-debug`

Associate it with dbg will make it unusable in a project that's large enough. It will eventually result in the linker error .bss truncated.

garymm:
I'm rather new to CUDA, but I can't seem to find anybody complaining of .bss truncated linker errors issue on https://forums.developer.nvidia.com/ or anywhere else. 

Also I just checked with LLVM 16, and it seems that by default clang will emit debug information for both host and device code when `-g` is set, so this change is also making the clang and nvcc toolchains to behave the same.
You have to pass separate flags to tell clang to disable debug info for the device code, according to [this discussion](https://discourse.llvm.org/t/debug-info-for-cuda-code/50473/27).

It's obviously your repo so you get to decide, but keeping this consistent between NVCC and clang and host and device and using the intuitively obvious flag seems better to me.

Let me know if I've convinced you :-)

cloudhan:
@garymm .bss truncated, for example, [pytorch/issues/57206](https://togithub.com/pytorch/pytorch/issues/57206). These things are rarely met, but once you face it, it will be extremely annoying and hard to workaround.

Additionally, the device debug impose huge perf panelty, which may not be desired by default, so there must be a way to disable it. As for clang, we better also disable the device debug symbol by default if possible.

garymm:
OK, see https://github.com/bazel-contrib/rules_cuda/pull/108

Issue #113, opened 2023-06-28
libcudart.so not found on system with multiple CUDA versions


garymm:
On a system with two versions of CUDA installed like so:
```
➜  ls -ld /usr/local/cuda*
lrwxrwxrwx  1 root   20 Jun 23 14:05 /usr/local/cuda -> /usr/local/cuda-11.8/
drwxr-xr-x 16 root 4096 Jun 23 13:54 /usr/local/cuda-11.8/
lrwxrwxrwx  1 root   25 Jun  9 18:58 /usr/local/cuda-12 -> /etc/alternatives/cuda-12/
drwxr-xr-x 15 root 4096 Jun 23 14:04 /usr/local/cuda-12.1/
```
And the following WORKSPACE.bazel code:

```bzl
load("@rules_cuda//cuda:repositories.bzl", "local_cuda", "register_detected_cuda_toolchains")

local_cuda(
    name = "local_cuda",
    toolkit_path = "/usr/local/cuda-12.1",
)

register_detected_cuda_toolchains()
```

Binaries seem to get linked against the wrong libcuda.so:
```
➜  ldd bazel-bin/foo | grep libcuda
	libcudart.so.12 => not found
```

On a system that only has one version of CUDA installed, with the same workspace snippet, it seems like libcudart.so is pointing to `/usr/local/cuda`, NOT the `/usr/local/cuda-12.1` specified in the WORKSPACE.

```
➜  ldd bazel-bin/foo | grep libcuda
	libcudart.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libcudart.so.12 (0x00007f2cdad20000)
```

Seems the linker is somehow being told to look at `/usr/local/cuda` regardless of the toolkit_path specified.

cloudhan:
It is because `cc_import`
https://github.com/bazel-contrib/rules_cuda/blob/18eba6d463db35253788a94fdaa45e9863ab19f0/cuda/runtime/BUILD.local_cuda#L42-L46
cause it linking with **rpath** ([by (my) design](https://github.com/bazel-contrib/rules_cuda/commit/7c88d0a994bef366a5c27bc7b717b060f5b7ba13)).

The problem is multi-fold here.

1. First, I'd assume that you are facing this problem because you are not run it with `bazel run` or `bazel test`.
2. Second question is, what is the rpath here? `patchelf --print-rpath bazel-bin/cublas/main` shows it is (manually splited due to long line) 
    ```
[...]

PR #114, opened 2023-06-29
Include versioned so files in cuda_runtime

garymm:
libcudart.so is a symlink to a versioned .so file, and the SONAME
specifies the versioned file, which means that at runtime the dynamic
linker will look for the versioned file, so we need to import the
versioned files.

Fixes: https://github.com/bazel-contrib/rules_cuda/issues/113

### DIFF
--- a/cuda/runtime/BUILD.local_cuda
+++ b/cuda/runtime/BUILD.local_cuda
@@ -1,5 +1,4 @@
-load("@rules_cuda//cuda:defs.bzl", "if_linux", "if_windows")
-load(":defs.bzl", "if_local_cuda")
+load("@rules_cuda//cuda:defs.bzl", "cc_import_versioned_sos", "if_linux", "if_windows")
 
 package(
     default_visibility = ["//visibility:public"],
@@ -8,9 +7,9 @@ package(
 filegroup(
     name = "compiler_deps",
     srcs = [
+        "cuda/version.txt",
         ":_cuda_header_files",
     ] + glob([
-        "cuda/version.txt",
         "cuda/bin/**",
         "cuda/lib64/**",
         "cuda/nvvm/**",
@@ -39,10 +38,9 @@ cc_library(
     ]),
 )
 
-cc_import(
+cc_import_versioned_sos(
     name = "cudart_so",
     shared_library = "cuda/lib64/libcudart.so",
-    target_compatible_with = ["@platforms//os:linux"],
 )
 
 cc_library(
@@ -136,16 +134,14 @@ cc_library(
     ]),
 )
 
-cc_import(
+cc_import_versioned_sos(
     name = "cublas_so",
     shared_library = "cuda/lib64/libcublas.so",
-    target_compatible_with = ["@platforms//os:linux"],
 )
 
-cc_import(
+cc_import_versioned_sos(
     name = "cublasLt_so",
     shared_library = "cuda/lib64/libcublasLt.so",
-    target_compatible_with = ["@platforms//os:linux"],
 )
 
 cc_import(
@@ -193,10 +189,9 @@ cc_library(
 )
 
 # CUPTI
-cc_import(
+cc_import_versioned_sos(
     name = "cupti_so",
     shared_library = "cuda/lib64/libcupti.so",
-    target_compatible_with = ["@platforms//os:linux"],
 )
 
 cc_import(
@@ -301,10 +296,9 @@ cc_library(
 )
 
 # curand
-cc_import(
+cc_import_versioned_sos(
     name = "curand_so",
     shared_library = "cuda/lib64/libcurand.so",
-    target_compatible_with = ["@platforms//os:linux"],
 )
 
 cc_import(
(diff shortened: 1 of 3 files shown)


garymm:
I'm guessing the checks failed due to a GitHub outage. Please retry them if you're able.

cuda/runtime/BUILD.local_cuda
```diff
@@ -39,10 +39,19 @@ cc_library(
...
+
+[cc_import(
+    name = paths.basename(p),
+    shared_library = p,
     target_compatible_with = ["@platforms//os:linux"],
+) for p in cudart_so_paths]
+
+cc_library(
+    name = "cudart_so",
+    deps = [":%s" % paths.basename(p) for p in cudart_so_paths],
```
cloudhan, line 54: Now it shows the following:
```
> ldd bazel-bin/basic/main
        linux-vdso.so.1 (0x00007ffff4bfd000)
        libcudart.so.12 => /home/cloud/rules_cuda/examples/bazel-bin/basic/../_solib_k8/_U@local_Ucuda_S_S_Clibcudart.so.12___Ucuda_Slib64/libcudart.so.12 (0x00007f0066c00000)
        ...
```

It starts to use the full mangled path again. This dismissed the purpose to use `cc_import`, acts the same as `cuda_library(srcs=[<lib_path>])`.

This is not something reasonable path outside of bazel environment, better avoid this case totally.

garymm: Thanks for taking a look.
Prior to this PR, I believe the way things are working is the following (please correct me if I'm wrong):
* `repositories.local_cuda()` creates a bazel repository which is a symlink to some absolute path, default  `/usr/local/cuda`.
* A `cuda_library` depends on `@local_cuda//:cuda_runtime`.
* In its sandbox, bazel creates a directory which only contains the declared dependencies and passes it to the linker with `-L`. In this case, that means it includes `libcudart.so`, but NOT `libcudart.so.12` (replace 12 with whatever version you want) is being used. You can see this in the linker params with a line like: `-Lbazel-out/k8-fastbuild/bin/_solib_k8/_U@local_Ucuda_S_S_Ccudart_Uso___Ucuda_Slib64`, and

```sh
➜  ls bazel-out/k8-fastbuild/bin/_solib_k8/_U@local_Ucuda_S_S_Ccudart_Uso___Ucuda_Slib64
libcudart.so@
```
* Similarly the linker specifies the `rpath` to include a directory inside `runfiles`, which also contains just libcdart.so: `-rpath $ORIGIN/main.runfiles/rules_cuda_examples/_solib_k8/_U@local_Ucuda_S_S_Ccudart_Uso___Ucuda_Slib64`.
* The linker opens `libcudart.so`, which is a symlink (possibly to another symlink). The final non-link target file has SONAME set to a versioned name like `libcudart.so.12`. So the binary is linked against `libcudart.so.12`.
* At runtime (including `bazel test`), the dynamic linker `ldd` will look in the runfiles for `libcudart.so.12` and fail to find it. Then it will look in system search paths. On my machine it will look in `/lib/x86_64-linux-gnu`, amongst other paths.
* If there is a libcudart.so.12 in any of the system paths, then it works. Else, it fails.

The situation that triggered the failure for me was:

1. Multiple versions of CUDA installed, 11 and 12.
2. `/lib/x86_64_linux-gnu/libcudart.so` is a symlink to `/lib/x86_64_linux-gnu/libcudart.so.11`.
3. In my WORKSPACE.bazel I have: `local_cuda(name = "local_cuda", toolkit_path = "/usr/local/cuda-12")`

In this case the binary gets linked against `libcudart.so.12` but the `rpath` does NOT contain `/usr/local/cuda-12`, and the system's default lib paths also don't contain it, so ldd fails to find it at runtime.

So I think this proposed change will not actually make things any worse as to your concern:

> This is not something reasonable path outside of bazel environment

The only way this works at runtime before this PR is if `ldd` searched the system library search paths. It will continue to do that in the case that the `libcudart.so.12` is missing from runfiles (I tested this by `rm`'ing the .so file from the runfiles and confirmed that `ldd` then finds the one that it used to find before this PR).

So I'm pretty sure this will not cause any regressions at all in things that currently work, but it will fix things that currently do not.

cloudhan: OK, my bad, it seem I am blind at some point. xxxD
```
> patchelf --print-needed bazel-bin/basic/main
libcudart.so.12
libstdc++.so.6
libm.so.6
libc.so.6
```

They are correctly resolved.

I am wondering is it willing for you to design a macro `get_versioned_so("cuda/lib64/libcudart.so", n=1)` and it should resolve to `"cuda/lib64/libcudart.so.12"` and `get_versioned_so("cuda/lib64/libcudart.so", n=2)` and resolve to `"cuda/lib64/libcudart.so.12.1"` so that we can apply it to other libs.

(So that I don't need to do it myself in the future.)

garymm: I added a macro and applied to tall of the libraries that appear to benefit from it. PTAL.

cloudhan (approved):
LGTM, thanks.

cloudhan:
@jsharpe Any concern for you?

jsharpe:
> @jsharpe Any concern for you?

No this looks good to me - in fact with the hermetic toolchain setup this is also an issue; I work around this in our internal build by fixing it up in a rules_pkg setup.

PR #118, opened 2023-07-01
Ignore MODULE.bazel for renovate bot

cloudhan:
In #117, bzlmod deps is auto updated but is not desired. Configure renovate bot to ignore `MODULE.bazel` scanning to avoid future auto updating.


cloudhan:
@jsharpe I think this is sufficient to avoid the undesired auto updating in future. I can also revert that change in this PR.

jsharpe (approved):

PR #120, opened 2023-07-06
set nvcc --generate-line-info when compilation_mode=fastbuild

garymm:
this makes the debug info for both host and device code the same
(i.e. --generate-line-info does for device code what -g1 does for
host code).

This is an intuitive result IMO.


jsharpe (approved):
LGTM

cloudhan:
@garymm Please update the testcase in https://github.com/bazel-contrib/rules_cuda/blob/main/tests/flag/BUILD.bazel to reflect the change.

garymm:
@cloudhan done

PR #123, opened 2023-07-06
add .bazel suffix to all BUILD files

garymm:
this makes syntax highlighting, code search etc easier and is consistent

Change-Id: Id1ce92b13f9a87b84ab513e36726428a61b97e6a


cuda/runtime/BUILD.local_cuda.bazel
cloudhan, line 1: I known this is weird, but please don't do this. The bazel community exhibit a weird practice here. And there is no established convention. So if it is not broken, don't fix it.

garymm: It is broken: syntax highlighting, buildifier, code search are all much harder without proper file extensions.

garymm:
Still valid but closing since it's been ignored for so long.

PR #125, opened 2023-07-06
have cuda_library output RDC

garymm:
This allows a `cuda_library` that was built with `rdc=True` to be
depended upon by another such library. This is convenient as such a
library can be consumed by either another `cuda_library` OR a
`cc_library`.

Note: if we have this, I'm not sure why we need a separate cuda_objects
rule anymore. Perhaps it can be removed in a later PR.

Change-Id: I1014d28a0ab3a9c76b788821211b13c4a9956d2a

### DIFF
--- a/cuda/private/rules/cuda_library.bzl
+++ b/cuda/private/rules/cuda_library.bzl
@@ -31,13 +31,17 @@ def _cuda_library_impl(ctx):
     # outputs
     objects = depset(compile(ctx, cuda_toolchain, cc_toolchain, src_files, common, pic = False, rdc = use_rdc))
     pic_objects = depset(compile(ctx, cuda_toolchain, cc_toolchain, src_files, common, pic = True, rdc = use_rdc))
+    rdc_objects = depset([])
+    rdc_pic_objects = depset([])
 
     # if rdc is enabled for this cuda_library, then we need futher do a pass of device link
     if use_rdc:
         transitive_objects = depset(transitive = [dep[CudaInfo].rdc_objects for dep in attr.deps if CudaInfo in dep])
         transitive_pic_objects = depset(transitive = [dep[CudaInfo].rdc_pic_objects for dep in attr.deps if CudaInfo in dep])
         objects = depset(transitive = [objects, transitive_objects])
+        rdc_objects = objects
         pic_objects = depset(transitive = [pic_objects, transitive_pic_objects])
+        rdc_pic_objects = pic_objects
         dlink_object = depset([device_link(ctx, cuda_toolchain, cc_toolchain, objects, common, pic = False, rdc = use_rdc)])
         dlink_pic_object = depset([device_link(ctx, cuda_toolchain, cc_toolchain, pic_objects, common, pic = True, rdc = use_rdc)])
         objects = depset(transitive = [objects, dlink_object])
@@ -87,12 +91,20 @@ def _cuda_library_impl(ctx):
             pic_lib = pic_libs,
             objects = objects,
             pic_objects = pic_objects,
+            rdc_objects = rdc_objects,
+            rdc_pic_objects = rdc_pic_objects,
         ),
         CcInfo(
             compilation_context = cc_info.compilation_context,
             linking_context = cc_info.linking_context,
         ),
-        cuda_helper.create_cuda_info(defines = depset(common.defines)),
+        cuda_helper.create_cuda_info(
+            defines = depset(common.defines),
+            objects = objects,
+            pic_objects = pic_objects,
+            rdc_objects = rdc_objects,
+            rdc_pic_objects = rdc_pic_objects,
+        ),
     ]
 
 cuda_library = rule(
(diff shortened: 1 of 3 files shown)


cloudhan:
This somehow allows cc_library to depend on an incomplete cuda_library, is there a way to avoid it?

garymm:
> This somehow allows cc_library to depend on an incomplete cuda_library, is there a way to avoid it?

Can you please elaborate? Or illustrate with an example? If `rdc=True`, the cuda_library will always perform a device link step, so it should be complete.

cuda/private/rules/cuda_library.bzl
```diff
@@ -87,12 +91,19 @@ def _cuda_library_impl(ctx):
             pic_lib = pic_libs,
             objects = objects,
             pic_objects = pic_objects,
+            rdc_objects = rdc_objects,
```
cloudhan, line 94: Are you missing `pic_rdc_objects` here?

cuda/private/rules/cuda_library.bzl
```diff
@@ -104,7 +115,7 @@ cuda_library = rule(
         "hdrs": attr.label_list(allow_files = ALLOW_CUDA_HDRS),
         "deps": attr.label_list(providers = [[CcInfo], [CudaInfo]]),
         "alwayslink": attr.bool(default = False),
-        "rdc": attr.bool(default = False, doc = "whether to perform relocateable device code linking, otherwise, normal device link."),
+        "rdc": attr.bool(default = False, doc = "Whether to produce and consume relocateable device code. Transitive deps that contain device code must all either be cuda_objects or cuda_library that has rdc = True. If False, all device code must be in the same translation unit. May have performance implications. See https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#using-separate-compilation-in-cuda."),
```
cloudhan, line 118: Could you please breakline a little bit, way too long for it.

cloudhan:
> If rdc=True, the cuda_library will always perform a device link step, so it should be complete.

Indeed.

The only limitation left is this cannot support circular object dependency due to the mandatory device link, which by itself is really stupid so I don't think user will need it generally.

cloudhan:
Backlog for memo here:

Originally, it is modeled as 

| host code | device code |
| --------- | ----------- | 
| cc_library for reusable cpu objects | cuda_objects for reusable device objects  |
| cc_binary for linked cpu executable | cuda_library for linked device executable |

For non-rdc and non-dlto case, the device executable is automatically linked already. `rdc` will enforce device_link then the compiler will ensure the device executable is linked.

This is something like `cc_binary` expose all objects for reusing with other `cc_binary`s. But this is not possible because the `main` entrypoint symbol conflict. For device case, the entrypoint is determined by a kernel launch, so this is not a problem.

Thus, allow `cuda_library` to expose the underlying objects for reusing between `cuda_library` but not `cc_library` will be harmless then.

cuda/private/rules/cuda_library.bzl
garymm: Done.

cuda/private/rules/cuda_library.bzl
garymm: Done.

cloudhan (approved):
Thanks for the improvement.

garymm:
@cloudhan would you be OK with me deleting cuda_objects? Seems useless now.

cloudhan:
No

garymm:
Why?

garymm:
Can you add an example that illustrates the problem?
I don't understand exactly how 1 or 2 can happen.

cloudhan:
1 is not very important, because those imcomplete objects are not exposed by `CcInfo`. For 2

https://github.com/bazel-contrib/rules_cuda/blob/22a46e6b3e6aaecd1fd4d3769a669d2861115068/cuda/private/rules/cuda_library.bzl#L38-L48

Lets consider the issue 162, build the shared library will cause

```
/usr/bin/ld.gold: error: bazel-out/k8-fastbuild/bin/tmp/_objs/mul/mul.rdc.pic.o: multiple definition of 'mul(double*, double const*, double const*, int)'
/usr/bin/ld.gold: bazel-out/k8-fastbuild/bin/tmp/_objs/mul/mul.rdc.pic.o: previous definition here
/usr/bin/ld.gold: error: bazel-out/k8-fastbuild/bin/tmp/_objs/mul/mul.rdc.pic.o: multiple definition of 'mul_impl(double*, double const*, double const*, int)'
/usr/bin/ld.gold: bazel-out/k8-fastbuild/bin/tmp/_objs/mul/mul.rdc.pic.o: previous definition here
/usr/bin/ld.gold: error: bazel-out/k8-fastbuild/bin/tmp/_objs/mul/mul.rdc.pic.o: multiple definition of '__fatbinwrap_4280babd_6_mul_cc_a09afee1'
/usr/bin/ld.gold: bazel-out/k8-fastbuild/bin/tmp/_objs/mul/mul.rdc.pic.o: previous definition here
/usr/bin/ld.gold: error: bazel-out/k8-fastbuild/bin/tmp/_objs/mul/mul.rdc.pic.o: multiple definition of '__device_stub__Z8mul_implPdPKdS1_i(double*, double const*, double const*, int)'
/usr/bin/ld.gold: bazel-out/k8-fastbuild/bin/tmp/_objs/mul/mul.rdc.pic.o: previous definition here
/usr/bin/ld.gold: error: bazel-out/k8-fastbuild/bin/tmp/_objs/mul/mul_dlink.rdc.pic.o: multiple definition of '__cudaRegisterLinkedBinary_4280babd_6_mul_cc_a09afee1'
/usr/bin/ld.gold: bazel-out/k8-fastbuild/bin/tmp/_objs/muladd/muladd_dlink.rdc.pic.o: previous definition here
```

```
| target | transitive_objects | objects (exposed via CcInfo) | rdc_objects (exposed via CudaInfo)
|--------|--------------------|------------------------------|------------------------------------
| mul    |                    | mul.o,mul_dlink.o            | mul.o
| muladd | mul.o              | muladd.o,mul.o,muladd_dlink.o| muladd.o,mul.o
```

And `cc_shared_library(name = "t", deps = [":mul", ":muladd"])` is valid because from user side, they want symbols in target `mul` to be included in target `t`, so they whole archive link `mul`, and this applies to `muladd` respectively.

cloudhan:
@garymm https://github.com/bazel-contrib/rules_cuda/issues/164#issuecomment-1721590261 This should address all the issue we are facing. Let me fix this.

PR #130, opened 2023-07-13
add support for `.optf` files

corin33:
# summary
Adds support for passing files via the `-optf` flag to `nvcc`, by means of a new attribute named `optf` in `cuda_objects()` and `cuda_library()`.

# why

Currently, the exact compiler flags need to be known prior to calling `bazel`, as they are plain string attributes.
There is no way to set them during analysis stage (say, to have them depend on the system architecture), except for the global `//cuda:copts`.
Adding support for the `-optf` flag of `nvcc` resolves this. The programmer can now have an `optf` file per library or per compilation unit.
One could also maintain multiple sets of files and choose between them during analysis stage (e.g. for loading optimized flags per target processor), or even generate the files during build time (using `genrule()`).
 
Also, passing required compiler arguments and defines via `.optf` files, instead of setting them in the `BUILD` file, is a great way to keep code and build system configuration separated.

# diff
* add `optf` attribute of label type `File` to `cuda_objects()` and `cuda_library()`
* add the file passed via the `optf` attribute as file dependency to the target of the compilation unit (object or library)
* add the file name of the `optf` file as `--options-file <filename>` to the compiler command line
* add an example


cloudhan:
> Currently, the exact compiler flags need to be known prior to calling `bazel`, as they are plain string attributes.
There is no way to set them during analysis stage (say, to have them depend on the system architecture), except for the global `//cuda:copts`.

Not true. You need to learn how to **correctly** use bazel first. See https://bazel.build/docs/configurable-attributes. Your optf also defeat the purpose of your claim, without using a `select`, you cannot depend optf on the system architecture either.

> Adding support for the -optf flag of nvcc resolves this. The programmer can now have an optf file per library or per compilation unit.

The copts can already do this, just assigning different set options as lists of strings solves the problem.

You also need to learn how to write `.bazelrc`. See https://bazel.build/run/bazelrc. Not in an imaginary way of yourself.

ryanleary:
Closing since this appears possible with copts. If further discussion needed, please reopen.

PR #132, opened 2023-07-16
The flags `-x c++` should be passed as `compile_flags`, instead of `host_compile_flags`

cloudhan:


cloudhan:
This is not necessary, only will be triggered if we force nvcc to use wrapper device link.

PR #134, opened 2023-07-17
Fix incorrect tools label for _wrapper_device_link

cloudhan:
Label, for example, `:cuda/bin/crt/link.stub` is incorrect in

https://github.com/bazel-contrib/rules_cuda/blob/3ce19bc587f5f4404402340e43a22fb192c17dac/cuda/private/repositories.bzl#L69-L76

Prior to #66 it was, `@local_cuda//:cuda/bin/crt/link.stub` which resolves to `<local_cuda>/cuda/bin/crt/link.stub` , but the generated BUILD file resides in `<local_cuda>/toolchain/BUILD`, so `:cuda/bin/crt/link.stub` actually resolves to `<local_cuda>/toolchain/cuda/bin/crt/link.stub` which does not exists.

### DIFF
--- a/cuda/private/repositories.bzl
+++ b/cuda/private/repositories.bzl
@@ -61,19 +61,19 @@ def detect_cuda_toolkit(repository_ctx):
         cuda_path = None
 
     bin_ext = ".exe" if _is_windows(repository_ctx) else ""
-    nvlink = "@rules_cuda//cuda/dummy:nvlink"
-    link_stub = "@rules_cuda//cuda/dummy:link.stub"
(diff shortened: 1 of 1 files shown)


cloudhan:
@jsharpe I think we should have a `label_helper` for creating `Label` of `@rules_cuda` and `@local_cuda` instead of `str`, so that bazel can help with the remapping.

Previous try of `str(Label("@rules_cuda//cuda/dummy:nvlink"))` caused the conversion back to `str` as `//cuda/dummy:nvlink` instead of the canonical form `@<rules_cuda_remapped_name>//cuda/dummy:nvlink` in the `rules_cuda` repo.

PR #135, opened 2023-07-19
Fix _wrapper_device_link artifact name conflict

cloudhan:
Compile the newly added example with `_wrapper_device_link` with rdc cause artifact name conflict

https://github.com/bazel-contrib/rules_cuda/blob/0503ead4724517db445796cbcdfc8400ce88c853/examples/rdc/BUILD.bazel#L39-L44

The reason is because our `compile` always compile to `<target_output_dir>/_objs`:

```
cuda_library(name="b") --> artifact base name is `b` during `compile`

with `compile` for following srcs
"b.cu" ------------------> artifact `<target_output_dir>/_objs/b.rdc.o`
`fatbin.cu` ------rdc----> artifact `<target_output_dir>/_objs/b.rdc.o`

After this PR:
"b.cu" ------------------> artifact `<target_output_dir>/_objs/b.rdc.o`
`fatbin.cu` ------rdc----> artifact `<target_output_dir>/_objs/_dlink/b.rdc.o`
```

Note we reserve the `_prefix` keyword params for future change and keep it as (implicitly) internal.

### DIFF
--- a/cuda/private/actions/compile.bzl
+++ b/cuda/private/actions/compile.bzl
@@ -9,7 +9,8 @@ def compile(
         srcs,
         common,
         pic = False,
-        rdc = False):
+        rdc = False,
+        _prefix = "_objs"):
     """Perform CUDA compilation, return compiled object files.
 
     Notes:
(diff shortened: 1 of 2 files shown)


cuda/private/actions/compile.bzl
```diff
@@ -9,7 +9,8 @@ def compile(
         srcs,
         common,
         pic = False,
-        rdc = False):
+        rdc = False,
+        _prefix = "_objs"):
```
jsharpe, line 13: Aren't you missing the change in this function to actually make use of this prefix?

cloudhan: Yes! Burried in another branch.

PR #144, opened 2023-08-22
Feat cuda version

haibozhang123:
support specaified CUDA Version

Issue:
In some cases, system have different version's CUDA toolkit ，so it need to specified CUDA version param（like Windows CI client-server ，and need to build history code to get stable unified products ）

so it to need to specified CUDA version
If special CUDA version , bazel test need to get special CUDA version's shared library runtime file ( dll file)
effect :
if not set cuda_version or dll_runtime_system_provided param (default none or True), it will be no impact
if set CUDA version and can not find cuda_version 's CUDA toolkit, it will find CUDA in system as old logic, it will be no impact


WORKSPACE.bazel
```diff
@@ -7,7 +7,7 @@ local_repository(
 
 load("//cuda:repositories.bzl", "register_detected_cuda_toolchains", "rules_cuda_dependencies")
 
-rules_cuda_dependencies()
+rules_cuda_dependencies("11.8")
```
jsharpe, line 10: These can't be hardcoded as this breaks the CI tests of different versions of CUDA

examples/WORKSPACE.bazel
```diff
@@ -14,6 +14,5 @@ local_repository(
 
 load("@rules_cuda//cuda:repositories.bzl", "register_detected_cuda_toolchains", "rules_cuda_dependencies")
 
-rules_cuda_dependencies()
-
+rules_cuda_dependencies("11.8")
```
jsharpe, line 17: As above

cuda/runtime/BUILD.local_cuda
```diff
@@ -52,14 +55,15 @@ cc_library(
...
-    system_provided = 1,
+    shared_library = "cuda/bin/cudart64_"+ str(cuda_package_version("cuda_cudart")) + ".dll",
     target_compatible_with = ["@platforms//os:windows"],
 )
 
 cc_import(
     name = "cudadevrt_lib",
     interface_library = "cuda/lib/x64/cudadevrt.lib",
     system_provided = 1,
+    #shared_library = "cuda/bin/cudart64_"+ str(cuda_package_version("cuda_cudart")) + ".dll",
```
jsharpe, line 66: Why is this one commented out?

cuda/runtime/BUILD.local_cuda
```diff
@@ -52,14 +55,15 @@ cc_library(
 cc_import(
     name = "cudart_lib",
     interface_library = "cuda/lib/x64/cudart.lib",
-    system_provided = 1,
+    shared_library = "cuda/bin/cudart64_"+ str(cuda_package_version("cuda_cudart")) + ".dll",
```
jsharpe, line 58: This seems like it might be x64 specific from the name? Do you need to also provide the 32 bit variants of these libraries or are they always only 64 bit? I'm not so familiar with cuda libraries on windows.
If its x64 only then the `target_compatible_with` option will also need updating.

cuda/private/repositories.bzl
```diff
@@ -232,4 +311,5 @@ def rules_cuda_dependencies(toolkit_path = None):
         ],
     )
 
-    local_cuda(name = "local_cuda", toolkit_path = toolkit_path)
+    local_cuda(name = "local_cuda", cuda_version = cuda_version, toolkit_path = toolkit_path)
```
jsharpe, line 314: There's similar calls to local_cuda that need updating in the bzlmod module extension in cuda/extensions.bzl. You'll need to expose this same cuda_version option in the modules tag_class.

cloudhan:
Why is this needed?

PR #147, opened 2023-08-29
Add attr provides for cuda_library like cc_library

### LINKED ISSUE #146: Add attr `provides` for cuda_library like cc_library
I need to use the `required_providers` in [aspect](https://bazel.build/rules/lib/globals/bzl.html#aspect) for specific filters.

But cuda_library doesn't offer the `provides` field [like the cc_library does](https://github.com/bazelbuild/bazel/blob/master/src/main/starlark/builtins_bzl/common/cc/cc_library.bzl#L627C5-L627C13).

See: https://github.com/bazelbuild/bazel/issues/17214#issuecomment-1429452869

Zeratul-Aiur:
Fixes #146.

### DIFF
--- a/cuda/private/rules/cuda_library.bzl
+++ b/cuda/private/rules/cuda_library.bzl
@@ -141,4 +141,5 @@ cuda_library = rule(
     },
     fragments = ["cpp"],
     toolchains = use_cpp_toolchain() + use_cuda_toolchain(),
+    provides = [CcInfo],
 )


cuda/private/rules/cuda_library.bzl
```diff
@@ -141,4 +141,5 @@ cuda_library = rule(
     },
     fragments = ["cpp"],
     toolchains = use_cpp_toolchain() + use_cuda_toolchain(),
+    provides = [CcInfo],
```
cloudhan, line 144: Is it better to provide all of them? That is `[DefaultInfo, OutputGroupInfo, CcInfo, CudaInfo]`?

cloudhan:
@Zeratul-Aiur Thanks for the change. Could you also apply the change `cuda_objects.bzl`?

You can leave alone the pre-commit.

cuda/private/rules/cuda_library.bzl
Zeratul-Aiur: Thx,done.

PR #158, opened 2023-09-04
Add integration test CI pipeline

cloudhan:


jsharpe (approved):

cloudhan:
@jsharpe @ryanleary Currently, the pipeline is passively trigger, either manually workflow dispatch or comment in PR with `/test`. Maybe make it automatically triggers on main branch push. But let's see how it goes.

cloudhan:
This pipeline still needs some improvement. Mainly to avoid the unconditionally triggering of the pipeline (and then they are all skipped)

PR #168, opened 2023-09-17
Add cuda_test rule

### LINKED ISSUE #166: Proposal to add a `cuda_test` rule
When using the `cc_test` rule, tests that are using cuda are compiled with `cc` compilers and not with `nvcc`. An easy solution for this is to provide a rule `cuda_test` that delegates everything to the `cuda_libary` and finally uses a `cc_test` depending on the cuda target. Something like 

```starlark
def cuda_test(name, srcs = [], ...):
    cuda_library_name = "_" + name

    cuda_library(
        name = cuda_library_name,
        srcs = srcs,
        ...
    )

    cc_test(
        name = name,
        deps = [cuda_library_name],
        ...
    )
```

hofbi:
Fixes #166

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,6 +6,7 @@ module(
 
 bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "platforms", version = "0.0.6")
+bazel_dep(name = "rules_cc", version = "0.0.8")
 
 toolchain = use_extension("@rules_cuda//cuda:extensions.bzl", "toolchain")
 use_repo(toolchain, "local_cuda")
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,7 +6,6 @@ module(
 
 bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "platforms", version = "0.0.6")
-bazel_dep(name = "rules_cc", version = "0.0.8")
 
 toolchain = use_extension("@rules_cuda//cuda:extensions.bzl", "toolchain")
 use_repo(toolchain, "local_cuda")
--- a/cuda/private/macros/cuda_test.bzl
+++ b/cuda/private/macros/cuda_test.bzl
@@ -9,6 +9,7 @@ def cuda_test(**attrs):
         srcs = getattr(attrs, "srcs", []),
         copts = getattr(attrs, "copts", []),
         deps = getattr(attrs, "deps", []),
+        tags = getattr(attrs, "tags", []),
         testonly = True,
     )
 
--- a/cuda/private/macros/cuda_test.bzl
+++ b/cuda/private/macros/cuda_test.bzl
@@ -1,4 +1,3 @@
-load("@rules_cc//cc:defs.bzl", "cc_test")
 load("//cuda/private:rules/cuda_library.bzl", "cuda_library")
 
 def cuda_test(
@@ -17,7 +16,7 @@ def cuda_test(
         deps = deps,
     )
 
-    cc_test(
+    native.cc_test(
         name = name,
         copts = copts,
         visibility = visibility,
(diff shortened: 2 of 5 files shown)


MODULE.bazel
```diff
@@ -6,6 +6,7 @@ module(
 
 bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "platforms", version = "0.0.6")
+bazel_dep(name = "rules_cc", version = "0.0.8")
```
cloudhan, line 9: You don't need this.

cuda/private/rules/cuda_test.bzl
```diff
@@ -0,0 +1,26 @@
...
+    cuda_library_name = "_" + name
+
+    cuda_library(
+        name = cuda_library_name,
+        srcs = srcs,
+        copts = copts,
+        deps = deps,
+    )
+
+    cc_test(
```
cloudhan, line 20: You can use native.cc_test

cuda/private/rules/cuda_test.bzl
cloudhan, line 1: Could you move this file to **macros**/cuda_test.bzl?

cuda/private/macros/cuda_test.bzl
```diff
@@ -0,0 +1,25 @@
+load("//cuda/private:rules/cuda_library.bzl", "cuda_library")
+
+def cuda_test(
+        name,
+        srcs = [],
+        copts = [],
+        visibility = ["//visibility:private"],
```
jsharpe, line 7: This is not respecting the package level default visibility, this should be unset if not specified.

cuda/private/macros/cuda_test.bzl
```diff
@@ -0,0 +1,25 @@
...
+        data = [],
+        deps = []):
+    cuda_library_name = "_" + name
+
+    cuda_library(
```
jsharpe, line 12: Make the visibility of this target private to hide the implementation detail?

cuda/private/macros/cuda_test.bzl
```diff
@@ -0,0 +1,25 @@
...
+        deps = deps,
+    )
+
+    native.cc_test(
+        name = name,
+        copts = copts,
+        visibility = visibility,
+        data = data,
+        deps = [cuda_library_name],
+    )
```
jsharpe, line 25: What about all the other attributes for cc_test? In particular exec_comaptible_with and exec_properties are important for selecting the correct execution platform as the user needs to ensure this is run on a platform that has a cuda compatible GPU present.
The easiest way to do this is to forward kwargs to this rule so you don't have to explicitly list all attributes.

cuda/private/macros/cuda_test.bzl
```diff
@@ -0,0 +1,25 @@
+load("//cuda/private:rules/cuda_library.bzl", "cuda_library")
+
+def cuda_test(
```
jsharpe, line 3: Can you add a doc string so that the docs can include usage for this macro?

cuda/private/macros/cuda_test.bzl
jsharpe: Also can this target be marked as test only?

jsharpe (changes requested):
Main change required is to pass through kwargs so that all attributes are supported. Can you also make sure that this is documented in the generated docs?

cloudhan:
To enable the docs gen, re-export the symbol like the following:
https://github.com/bazel-contrib/rules_cuda/blob/4c0020c4fa05500d83dd93583fbd12c16d0d1b79/docs/user_docs.bzl#L5-L6

cuda/private/macros/cuda_test.bzl
```diff
@@ -0,0 +1,18 @@
...
+
+def cuda_test(**attrs):
+    """Wrapper to ensure the test is compiled with the CUDA compiler."""
+    cuda_library_name = "_" + getattr(attrs, "name", "")
+
+    _cuda_library(
+        name = cuda_library_name,
+        srcs = getattr(attrs, "srcs", []),
+        copts = getattr(attrs, "copts", []),
+        deps = getattr(attrs, "deps", []),
```
jsharpe, line 11: I think we also want to propagate tags into the _cuda_library?

jsharpe (commented):
LGTM - just one last change to propagate tags into the _cuda_library and then I think this is good to go.

PR #176, opened 2023-10-02
Fix cuda test duplicate deps attribute

hofbi:
* I realized that the previos impl contains deps twice in the cc_test which will be fixed here
* I also pass all attrs to the cuda_library


cloudhan (approved):
LGTM

PR #185, opened 2023-10-27
Pass through `--sysroot` to host compiler

lalten:
nvcc shells out to the host compiler, but does not set the `--sysroot` option. In the case of a hermetic cc_toolchain the host compiler will then fail to find the system includes.

See https://github.com/lalten/mre_rules_cuda-sysroot/tree/main for a demonstration

See discussion on Bazel Slack: https://bazelbuild.slack.com/archives/C043E282KUG/p1698396425839079

### DIFF
--- a/cuda/private/cuda_helper.bzl
+++ b/cuda/private/cuda_helper.bzl
@@ -3,6 +3,7 @@
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_skylib//lib:types.bzl", "types")
 load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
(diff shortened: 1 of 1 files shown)


cloudhan (approved):
LGTM. Thanks for the contribution.

Note we might need move it to feature based configuration in the future if some host compilers use different option name for sysroot, for now, lets keep it simple and stupid.

hypdeb:
> Note we might need move it to feature based configuration in the future if some host compilers use different option name for sysroot, for now, lets keep it simple and stupid.

Does that mean doing something like [this](https://github.com/bazelbuild/bazel/blob/master/tools/cpp/cc_flags_supplier_lib.bzl#L56) and pass all the flags that are passed to the host compiler through `-Xcompiler` ? If yes, I'll do it, because I think it is what is blocking me currently.

PR #192, opened 2023-11-02
Move non glob files out of glob

hofbi:


jsharpe (approved):

jsharpe:
Hmm, looks like the reason they're in the glob is to make them optional as they don't appear to be present in CI?

hofbi:
Right, but this causes issues with `--incompatible_disallow_empty_glob` (at least in my case)

cloudhan:
version.json and version.txt only exisits in specific versions, exclusively.

hofbi:
Then should we better go for `allow_empty = True` in the `glob` function?

cloudhan:
@hofbi Do we really need to set all globs to allow_empty=True?

hofbi:
I just did it for all to avoid issues with `--incompatible_disallow_empty_glob` that are different than my own

cloudhan (approved):

PR #8, opened 2022-07-05
eliminate cpu architecture constraint for nvcc

ryanleary:
This PR simply removes the `@platforms//cpu:x86_64` constraint from the nvcc toolchain template. I'm not sure if this is the preferred way to achieve this, but CUDA on linux is compatible with aarch64 for both servers and embedded use cases. This enables builds with on these platforms as well.

PR #208, opened 2024-01-03
eliminate cpu architecture constraint for clang

dmellosanjay:
This allows building with clang on aarch64 machines. Same as how https://github.com/bazel-contrib/rules_cuda/pull/8 did it for nvcc.


cloudhan (approved):
Thanks for the contribution!

PR #209, opened 2024-01-03
Fix cquery-ing with cuda targets

mvukov:
This makes possible to do the following:
```sh
cd examples
bazel cquery //if_cuda/... --@rules_cuda//cuda:enable=False
```
when one wants to run cquery (e.g. within [bazel-diff](https://github.com/Tinder/bazel-diff)) in an env without cuda (e.g. a CI).

I read the readme of if_cuda example. It's advertised that it's OK that bazel errors out if one want to build a cuda target when rules_cuda is disabled. I'd say that's a drastic measure. Folks should IMO mark their bazel targets like I did in this PR.

### DIFF
--- a/cuda/BUILD.bazel
+++ b/cuda/BUILD.bazel
@@ -34,8 +34,8 @@ config_setting(
 )
 
 config_setting(
-    name = "is_cuda_found",
-    flag_values = {"@local_cuda//:cuda_found": "True"},
+    name = "is_valid_toolchain_found",
+    flag_values = {"@local_cuda//:valid_toolchain_found": "True"},
 )
 
 # Command line flag to specify the list of CUDA architectures to compile for.
@@ -92,16 +92,16 @@ label_flag(
     build_setting_default = "@local_cuda//:cuda_runtime",
 )
 
-constraint_setting(name = "cuda_must_be_enabled_setting")
+constraint_setting(name = "rules_are_enabled_setting")
 
 constraint_value(
-    name = "cuda_must_be_enabled",
-    constraint_setting = ":cuda_must_be_enabled_setting",
+    name = "rules_are_enabled",
+    constraint_setting = ":rules_are_enabled_setting",
 )
 
-constraint_setting(name = "cuda_must_be_found_setting")
+constraint_setting(name = "valid_toolchain_is_configured_setting")
 
 constraint_value(
-    name = "cuda_must_be_found",
-    constraint_setting = ":cuda_must_be_found_setting",
+    name = "valid_toolchain_is_configured",
+    constraint_setting = ":valid_toolchain_is_configured_setting",
 )
--- a/cuda/BUILD.bazel
+++ b/cuda/BUILD.bazel
@@ -4,7 +4,6 @@ load(
     "bool_flag",
     "string_flag",
 )
-load("@bazel_skylib//lib:selects.bzl", "selects")
 load("//cuda/private:rules/flags.bzl", "cuda_archs_flag", "repeatable_string_flag")
 
 package(default_visibility = ["//visibility:public"])
@@ -37,14 +36,6 @@ config_setting(
 config_setting(
     name = "is_cuda_found",
     flag_values = {"@local_cuda//:cuda_found": "True"},
-)
-
-selects.config_setting_group(
-    name = "is_enabled_and_cuda_found",
-    match_all = [
-        ":is_enabled",
-        ":is_cuda_found",
-    ],
 )
 
 # Command line flag to specify the list of CUDA architectures to compile for.
@@ -99,4 +90,18 @@ repeatable_string_flag(
 label_flag(
     name = "runtime",
     build_setting_default = "@local_cuda//:cuda_runtime",
+)
+
+constraint_setting(name = "cuda_must_be_enabled_setting")
+
+constraint_value(
+    name = "cuda_must_be_enabled",
+    constraint_setting = ":cuda_must_be_enabled_setting",
+)
+
+constraint_setting(name = "cuda_must_be_found_setting")
+
+constraint_value(
+    name = "cuda_must_be_found",
+    constraint_setting = ":cuda_must_be_found_setting",
 )
--- a/cuda/BUILD.bazel
+++ b/cuda/BUILD.bazel
@@ -4,6 +4,7 @@ load(
     "bool_flag",
     "string_flag",
 )
+load("@bazel_skylib//lib:selects.bzl", "selects")
 load("//cuda/private:rules/flags.bzl", "cuda_archs_flag", "repeatable_string_flag")
 
 package(default_visibility = ["//visibility:public"])
@@ -31,6 +32,19 @@ bool_flag(
 config_setting(
     name = "is_enabled",
     flag_values = {":enable": "True"},
+)
+
+config_setting(
+    name = "is_cuda_found",
+    flag_values = {"@local_cuda//:cuda_found": "True"},
+)
+
+selects.config_setting_group(
+    name = "is_enabled_and_cuda_found",
+    match_all = [
+        ":is_enabled",
+        ":is_cuda_found",
+    ],
 )
 
 # Command line flag to specify the list of CUDA architectures to compile for.
--- a/cuda/private/defs.bzl
+++ b/cuda/private/defs.bzl
@@ -1,0 +1,16 @@
+"""private defs"""
+
+def requires_cuda():
+    """Returns constraint_setting that is satisfied if:
+
+    * CUDA is enabled and
+    * CUDA toolchain is found.
+
+    Add to 'target_compatible_with' attribute to mark a target incompatible when
+    @rules_cuda//cuda:is_enabled_and_cuda_found is not set. Incompatible targets are excluded
+    from bazel target wildcards and fail to build if requested explicitly.
+    """
+    return select({
+        "@rules_cuda//cuda:is_enabled_and_cuda_found": [],
+        "//conditions:default": ["@platforms//:incompatible"],
+    })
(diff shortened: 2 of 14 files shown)


cuda/private/toolchain_configs/dummy.bzl
```diff
@@ -0,0 +1,4 @@
+def _dummy_toolchain_config_impl(_ctx):
```
cloudhan, line 1: Could you please rename the "dummy" to something like "disable" (also the filename). So that the purpose is made clear.

cloudhan (changes requested):
Otherwise, looks good to me.

cloudhan:
It somehow don't work with bzlmod:

```console
> bazel cquery //if_cuda/... --@rules_cuda//cuda:enable=False --enable_bzlmod
ERROR: no such package '@@[unknown repo 'platforms' requested from @@]//': The repository '@@[unknown repo 'platforms' requested from @@]' could not be resolved: No repository visible as '@platforms' from main repository
ERROR: Analysis of target '//if_cuda:kernel' failed; build aborted: no such package '@@[unknown repo 'platforms' requested from @@]//': The repository '@@[unknown repo 'platforms' requested from @@]' could not be resolved: No repository visible as '@platforms' from main repository
ERROR: Build did NOT complete successfully
```

cuda/private/toolchain_configs/dummy.bzl
mvukov: Fixed.

mvukov:
* Fixed bzlmod for the disabled toolchain.
* Updated docs

Also, handled the case when cuda is enabled but cuda toolkit cannot be found. The previous behavior was that bazel build would fail because some targets are not present in `@local_cuda//*`, and thats true because @local_cuda//:BUILD was empty. This was not quite obvious/ergonomic unless you would dig into the code and found out that the BUILD file was empty. I refactored the code such that when we want to build a cuda target as e.g.

```
bazel build //if_cuda:kernel --@rules_cuda//cuda:enable=true
```
but there is no cuda toolkit available we get
```
ERROR: Target //if_cuda:kernel is incompatible and cannot be built, but was explicitly requested.
Dependency chain:
    //if_cuda:kernel (9867ef)   <-- target platform (@local_config_platform//:host) didn't satisfy constraint @platforms//:incompatible
FAILED: Build did NOT complete successfully (37 packages loaded, 138 targets configured)
```
This is IMO more readable, with addition of the docs for `requires_cuda` helper.

Thinking a bit further, it makes more sense to use globally `@rules_cuda//cuda:is_enabled_and_cuda_found` instead of `@rules_cuda//cuda:is_enabled`. But I'd like to hear your opinion on this. I'm open for different naming of `is_enabled_and_cuda_found` :)

mvukov:
@cloudhan PTAL.

cuda/private/defs.bzl
```diff
@@ -0,0 +1,16 @@
+"""private defs"""
+
+def requires_cuda():
```
mvukov, line 3: Taken and adapted from https://github.com/tensorflow/runtime, there used to be rules_cuda repo in there.

examples/if_cuda/BUILD.bazel
```diff
@@ -1,9 +1,10 @@
-load("@rules_cuda//cuda:defs.bzl", "cuda_library")
+load("@rules_cuda//cuda:defs.bzl", "cuda_library", "requires_cuda")
 
 cuda_library(
     name = "kernel",
     srcs = ["kernel.cu"],
     hdrs = ["kernel.h"],
+    target_compatible_with = requires_cuda(),
```
mvukov, line 7: Just a side note: users could still get non-so-obvious errors from bazel in case the target compatibility is not set.

cuda/private/defs.bzl
```diff
@@ -0,0 +1,16 @@
...
+    * CUDA is enabled and
+    * CUDA toolchain is found.
+
+    Add to 'target_compatible_with' attribute to mark a target incompatible when
+    @rules_cuda//cuda:is_enabled_and_cuda_found is not set. Incompatible targets are excluded
+    from bazel target wildcards and fail to build if requested explicitly.
+    """
+    return select({
+        "@rules_cuda//cuda:is_enabled_and_cuda_found": [],
+        "//conditions:default": ["@platforms//:incompatible"],
```
cloudhan, line 15: I am wondering if it is possible to alias `incompatible`, so that when `is_enabled` is false, it can somehow reports `incompatible` because the rule is disabled and when `is_cuda_found` is false, it can also report the toolchain is not configure.

Otherwise, when `incompatible` occurs, you still need to dig into it to figure out why it happens, and somehow make it a little bit worse because previous they are two distinct error, now they are just `incompatible`.

cloudhan:
> Thinking a bit further, it makes more sense to use globally @rules_cuda//cuda:is_enabled_and_cuda_found instead of @rules_cuda//cuda:is_enabled. But I'd like to hear your opinion on this.

I'd prefer not to add it. Because it somehow to me, it mixed the [**target** and **host**](https://bazel.build/extending/platforms), the **enable** generally means the `bazel build` invocation generate targets that is able to run cuda code (so you enable it). The toolchain part, however, should be regarded as the **host** where bazel runs is able to produce that target.

In the case of the toolchain being mis-configured or not configured, it is a `host_compatiable_with` thing (the attribute does not exist!).

@jsharpe Any comment?

cuda/private/defs.bzl
mvukov: Refactored.

mvukov:
I looked into how `"@platforms//:incompatible"` is defined and made a change in [d72a951](https://github.com/bazel-contrib/rules_cuda/pull/209/commits/d72a951b7ee1e8a924f94016ceb8fb1ff63f4969).

```
bazel build //if_cuda:kernel   --@rules_cuda//cuda:enable=false
ERROR: Target //if_cuda:kernel is incompatible and cannot be built, but was explicitly requested.
Dependency chain:
    //if_cuda:kernel (6b3a99)   <-- target platform (@local_config_platform//:host) didn't satisfy constraints [@rules_cuda//cuda:cuda_must_be_enabled, @rules_cuda//cuda:cuda_must_be_found]
FAILED: Build did NOT complete successfully (35 packages loaded, 139 targets configured)
```
The second constraint is a bit misleading because we didn't even look for it. Open to suggestions.
```
bazel build //if_cuda:kernel   --@rules_cuda//cuda:enable=true
ERROR: Target //if_cuda:kernel is incompatible and cannot be built, but was explicitly requested.
Dependency chain:
    //if_cuda:kernel (9867ef)   <-- target platform (@local_config_platform//:host) didn't satisfy constraint @rules_cuda//cuda:cuda_must_be_found
FAILED: Build did NOT complete successfully (0 packages loaded, 137 targets configured)
```

@cloudhan PTAL.

mvukov:
@cloudhan friendly reminder: PTAL. I implemented changes you requested.

cloudhan:
Sorry for the delay, was dealing with the CI update for Bazel 7 support.

cloudhan:
Let me play around with it to see if we can simplify it a little bit.

cloudhan:
To be honest, what is dragging me down in these changes is the latter maneuver that tries to hide error info
```
... No matching toolchains found for types @rules_cuda//cuda:toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@rules_cuda//cuda:toolchain_type'
```

It doesn't change the situation (that toolchain is not found), but add another layer of indirection.

mvukov:
Are you not happy about `requires_cuda()` macro logic but all the rest seems OK? If so, I can remove that macro and I'll have that in my monorepo.

cloudhan:
The `_requires_cuda_found()` part (aka, the constraint on cuda toolkit is found) . It hides the unresolved toolchain error with a custom constraint. Thus the builtin `No matching toolchains found for types @rules_cuda//cuda:toolchain_type` error is substituted with `didn't satisfy constraints ... @rules_cuda//cuda:cuda_must_be_found`

Better split it into a seperate PR? I might ask some bazel toolchain experts on what is the best practice for it...

cloudhan:
OK, I see your point now. The situation is cause by the ":disabled-local-toolchain" being relying on the `target_settings = [":cuda_is_disabled"],`. I tried to make it always available as a bottom case, but then the toolchain will ask for a `compiler_executable` from the toolchain config. 

The `compiler_executable` a string (mirrored from `CcToolchainInfo`), whereas it should have been a `Label` so that we can build a dummy one from `dummy.cc`, I am not sure if this is possible tho...
 
Lets adopt the current change to fix the issue. Please address naming tho.

cuda/BUILD.bazel
```diff
@@ -86,3 +91,17 @@ label_flag(
...
+    name = "cuda_must_be_enabled",
+    constraint_setting = ":cuda_must_be_enabled_setting",
+)
+
+constraint_setting(name = "cuda_must_be_found_setting")
+
+constraint_value(
+    name = "cuda_must_be_found",
+    constraint_setting = ":cuda_must_be_found_setting",
+)
```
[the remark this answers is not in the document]
mvukov: Fixed naming.

cuda/private/defs.bzl
```diff
@@ -0,0 +1,25 @@
...
+    """Returns constraint_setting that is satisfied if:
+
+    * CUDA is enabled and
+    * CUDA toolchain is found.
+
+    Add to 'target_compatible_with' attribute to mark a target incompatible when
+    the conditions are not satisfied. Incompatible targets are excluded
+    from bazel target wildcards and fail to build if requested explicitly.
+    """
+    return _requires_is_enabled() + _requires_cuda_found()
```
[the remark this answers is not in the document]
mvukov: Fixed.

cloudhan (approved):

PR #210, opened 2024-01-04
Add alwayslink to cuda_binary and cuda_test macros

### LINKED ISSUE #202: cuda_test missing alwayslink
For the reason discussed in https://github.com/bazel-contrib/rules_cuda/issues/20

`_cuda_library` inside `cuda_test` macro needs `alwayslink=True`. I don't know if there are side effects to set it by default. So filed an issue instead of a PR to add a line.

cloudhan:
Fix #202.

@skyw

### DIFF
--- a/cuda/private/macros/cuda_binary.bzl
+++ b/cuda/private/macros/cuda_binary.bzl
@@ -1,17 +1,29 @@
 load("//cuda/private:rules/cuda_library.bzl", _cuda_library = "cuda_library")
 
 def cuda_binary(name, **attrs):
-    """Wrapper to ensure the binary is compiled with the CUDA compiler."""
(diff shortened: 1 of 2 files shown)


skyw:
LGTM. Thanks for adding this!

ryanleary:
This looks good to me. Do we want to add some build test that verifies fix and prevents regression?

cloudhan:
It is manually verified and will be rarely changed. So let me merge this for now.

As for test, I am wondering how to verify it at this level without complicate the test itself too much. It is definitely not as easy as https://github.com/bazelbuild/bazel/blob/4603a3a76547e0d555b38372b2e060f6695cdf7c/src/test/java/com/google/devtools/build/lib/rules/cpp/StarlarkCcCommonTest.java#L5651-L5672

PR #215, opened 2024-01-16
Add additional tests for LTS releases

cloudhan:


.github/workflows/build-tests.yaml
```diff
@@ -68,3 +71,36 @@ jobs:
...
+      - run: echo "USE_BAZEL_VERSION=5.4.1" >> $GITHUB_ENV
+        if: ${{ !startsWith(matrix.cases.os, 'windows') }}
+      - run: echo "USE_BAZEL_VERSION=5.4.1" >> $env:GITHUB_ENV
+        if: ${{ startsWith(matrix.cases.os, 'windows') }}
+
+      - run: bazelisk build @rules_cuda_examples//basic:all
+      - run: bazelisk build @rules_cuda_examples//rdc:all
+      - run: bazelisk build @rules_cuda_examples//if_cuda:main
+      - run: bazelisk build @rules_cuda_examples//if_cuda:main --enable_cuda=False
+      # bzlmod is not covered, our separate @rules_cuda_examples repo setup doesn't play well with it
```
jsharpe, line 101: I think this is perfectly valid - bazel 5 really didn't support all the required bzlmod features needed for a functional bzlmod setup.

jsharpe (approved):

.github/workflows/build-tests.yaml
cloudhan: Right, it is perfectly valid as we have been exercising the practice for very long time without problem.

It is purely caused by the experimental feature does not correctly handle the repo setup at that time.

PR #222, opened 2024-02-03
Fix a typo in `if_cuda` doc

rygx:
The doc "bazel build //if_cuda:main --@rules_cuda//cuda:enable`" does not seem right.


cloudhan:
Thanks for pointing out!

PR #224, opened 2024-02-04
Ignore MODULE.bazel.lock file

### LINKED ISSUE #223: Untracked `MODULE.bazel.lock` files
bzlmod will automatically generate `MODULE.bazel.lock` files, but this file is neither tracked by git nor placed in .gitignore file, causing every time git to alert `MODULE.bazel.lock` not being tracked, and adds to chance that this file is committed inadvertently. IMO it could be a quite simple fix by choosing either of the options:
- Option 1: check in `MODULE.bazel.lock` files, if we are sure that this file is consistent across different machines (theoretically the lock file should maintain consistent on different machines, but I am not sure if there are any non-hermetic factors would cause environment-dependent contents).
- Option 2: add `MODULE.bazel.lock` file into `.gitignore`, if we are unsure if the lock file would be consistent. 

I would be happy to provide PR if either of the above options is a way to move forward.

rygx:
`MODULE.bazel.lock` file is a mechanism under `bzlmod` intended to record package and dependency versions (see Bazel documentation https://bazel.build/external/lockfile). However, this mechanism is not yet stable, for example, it does not seem to guarantee content consistency, and causes version control conflicts, for example see https://github.com/bazelbuild/bazel/issues/20369

So at this moment we put `MODULE.bazel.lock` files (including root directory and subdirectories) into `.gitignore` file.

Resolves #223

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,7 @@
 bazel-*
+
+# MODULE.bazel.lock is a new feature in bazel and is not stable yet.
+# Per https://github.com/bazelbuild/bazel/issues/20369, we ignore this file until 
+# MODULE.bazel.lock mechanism is stable and guarantee consistency across different 
+# development environments
+MODULE.bazel.lock


cloudhan:
It is hilarious to see we start to observe nvcc /tmp race condition in the CI. Because we have 4 cores for the free CI now 🤣

PR #239, opened 2024-04-19
feat: allow non-root module to register local_cuda

### LINKED ISSUE #238: How to use cublas in a non-root bazel module?
I'm using `rules_cuda` in a bazel MODULE `A`, and some of my `cuda_library` needs to link with `-lcubas` and `-lcublasLt`.

Naturally, I'm defining `local_cuda` like [examples/cublas/BUILD.bazel. ](https://github.com/bazel-contrib/rules_cuda/blob/main/examples/cublas/BUILD.bazel). In my `MODULE.bazel` of module `A`:

```bazel
bazel_dep(
    name = "rules_cuda",
    version = "0.2.1",
)

cuda.local_toolchain(
    name = "local_cuda",
    toolkit_path = "",
)
use_repo(cuda, "local_cuda")
```

And in my `BUILD.bazel`, I have the `cuda_library` target:

```bazel
cuda_library(
    name = ...,
    srcs = ...,
    deps = [
       "@local_cuda//:cublas",
    ],
    ...
)

```

This works fine when I build my module `A`. However, when I have another bazel module `B` that depends on module `A`, I cannot build module `B`, because the `local_cuda` can only be declared in a root module. I got this error:

> ERROR: Traceback (most recent call last):
>        File "/private/var/tmp/_bazel_dev/67de6cda420db4eb86e6ad3f1fd2b6e4/external/rules_cuda~/cuda/extensions.bzl", line 15, > column 21, in _init
>                fail("Only the root module may override the path for the local cuda toolchain")
> Error in fail: Only the root module may override the path for the local cuda toolchain

This is from this line: https://github.com/bazel-contrib/rules_cuda/blob/8f2f2e6d64d38e46d09538c921304c7c902a2564/cuda/extensions.bzl#L15. 

Is it possible to use `rules_cuda` in a bazel module that other modules can depend? I don't really need to customize my cuda path, as the default path works fine with me. Is there a way to avoid the above error?

cloudhan:
Potentially fix #238

### DIFF
--- a/cuda/extensions.bzl
+++ b/cuda/extensions.bzl
@@ -9,16 +9,22 @@ cuda_toolkit = tag_class(attrs = {
 
 def _init(module_ctx):
     registrations = {}
+    # first pass only handles root module
     for mod in module_ctx.modules:
+        if not mod.is_root:
+            continue
         for toolchain in mod.tags.local_toolchain:
-            if not mod.is_root:
-                fail("Only the root module may override the path for the local cuda toolchain")
             if toolchain.name in registrations.keys():
-                if toolchain.toolkit_path == registrations[toolchain.name]:
-                    # No problem to register a matching toolchain twice
-                    continue
-                fail("Multiple conflicting toolchains declared for name {} ({} and {}".format(toolchain.name, toolchain.toolkit_path, registrations[toolchain.name]))
+                fail("Multiple toolchains defined with the name \"{}\" during initialization of module \"{}\"".format(toolchain.name, mod.name))
             else:
+                registrations[toolchain.name] = toolchain.toolkit_path
+    # second pass only handles non-root module
+    for mod in module_ctx.modules:
+        if mod.is_root:
+            continue
+        for toolchain in mod.tags.local_toolchain:
+            # root module supersede non-root modules
+            if toolchain.name not in registrations.keys():
                 registrations[toolchain.name] = toolchain.toolkit_path
     for name, toolkit_path in registrations.items():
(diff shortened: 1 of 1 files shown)


cloudhan:
@jsharpe I am not quite remember why it was designed as previous. What is the concern to disallow non-root module to call `local_toolchain`? I'd like to verify it a little bit before adding tests.

jsharpe:
A non-root module can currently do this:
```
use_repo(cuda, "local_cuda")
```

This makes `local_cuda` visible in the scope of the non-root module but leaves the configuration of `local_cuda` still to be the root module as only the root module can tell you the path to where the cuda toolkit is installed.
I think the real problem though is the need for non-root modules to reach into `local_cuda` - in the future a hermetic toolchain would not be called `local_cuda` and so really we need another location that is independent of the configured toolchain to provide e.g. the `cublas` targets.

cloudhan:
It is not about cublas only, but rules instead. if you think abount it a bit further, say `cukernels` target is wrapped in a `clibrary` target and then is exposed as a public interface/target then all `clibrary` consumers will be required to configure rules_cuda explicitly. This will be quite annoying from the users' PoV. And it is contagious that makes it even worse.

`@local_cuda` is a bit misleading if we introduce hermetic toolchain, we still need sth like `@cuda_toolkit` to the host toolchains so that it can do the compilation (headers) and linking (libs) with the necessary files.

To me, it is not the configuration being a problem, someone need to do it at some level/point, it the endless propagation to downstream module/repo transitively being a problem.

PR #250, opened 2024-06-09
Fix spelling

Vertexwahn:
Fix spelling


cloudhan (approved):
Thanks for the fixing.

PR #259, opened 2024-07-12
Make it clear that "integrity" field expects SRI

zifeitong:
Documented in https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity


jsharpe (approved):
LGTM

cloudhan:
I don't think this is a related thing at all. The page referenced is a doc for web developers. The SRI suggested is for validating resources embedded in the html, it just happens to have common meaning with the checksum. This change actually look really bad to me.

cloudhan:
@zifeitong Thanks for pointing out. I am OK with either format. Simply put `sha256` is just some old boring format used in practice.

PR #262, opened 2024-08-07
ci: cover major bazel releases in utilities tests

cloudhan:
This reveal the release pipeline failure


cloudhan:
This should break current code base.

PR #263, opened 2024-08-07
test: workaround label resolving with bzlmod

cloudhan:
Partially fix test failures triggered by #262

Workaround utilities test when bzlmod is enabled:
```
ERROR: Traceback (most recent call last):
	File "/home/runner/work/rules_cuda/rules_cuda/tests/flag/flag_validation_test.bzl", line 98, column 61, in <toplevel>
		cuda_library_sm61_flag_test = _create_cuda_library_flag_test(config_settings_sm61)
	File "/home/runner/work/rules_cuda/rules_cuda/tests/flag/flag_validation_test.bzl", line 63, column 29, in _create_cuda_library_flag_test
		return analysistest.make(
	File "/home/runner/.cache/bazel/_bazel_runner/c7158fce5becee297c399541ca8c3db4/external/bazel_skylib~/lib/unittest.bzl", line [28](https://github.com/bazel-contrib/rules_cuda/actions/runs/10286317527/job/28467427384?pr=262#step:6:29)5, column 51, in _make_analysis_test
		test_transition = analysis_test_transition(
Error in analysis_test_transition: invalid transition output '@@[unknown repo '' requested from @@bazel_skylib~]//cuda:archs': no repo visible as @ from repository '@@bazel_skylib~'
ERROR: error loading package under directory 'tests': error loading package 'tests/flag': initialization of module 'tests/flag/flag_validation_test.bzl' failed
```

PR #264, opened 2024-08-07
fix(bzlmod): allow both root module and our module to call cuda.local_toolchain

### LINKED ISSUE #238: How to use cublas in a non-root bazel module?

cloudhan:
Superserde #239 

In https://github.com/bazel-contrib/rules_cuda/pull/263#issuecomment-2273722073 a new error popup as expected.

It is caused by the unittest secretly setup our module (rules_cuda, current root module as of invocation of `bazel` command) as an external module under testing, then the `local_cuda` is ignore in this case. We shall allow call to module_extenion in both root module and our module.

This in principle also fixes #238.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -8,6 +8,10 @@ bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "platforms", version = "0.0.6")
 
 cuda = use_extension("@rules_cuda//cuda:extensions.bzl", "toolchain")
+cuda.local_toolchain(
+    name = "local_cuda",
+    toolkit_path = "",
+)
 use_repo(cuda, "local_cuda")
 
 register_toolchains(
--- a/.github/workflows/build-tests.yaml
+++ b/.github/workflows/build-tests.yaml
@@ -71,6 +71,8 @@ jobs:
       - run: cd examples && bazelisk build --jobs=1 //if_cuda:main
       - run: cd examples && bazelisk build --jobs=1 //if_cuda:main --enable_cuda=False
       - run: bazelisk shutdown
+      # run some repo integration tests
+      - run: cd tests/integration && ./test_all.sh
 
       # Use Bazel 6
       - run: echo "USE_BAZEL_VERSION=6.4.0" >> $GITHUB_ENV
--- a/cuda/extensions.bzl
+++ b/cuda/extensions.bzl
@@ -7,19 +7,35 @@ cuda_toolkit = tag_class(attrs = {
     "toolkit_path": attr.string(doc = "Path to the CUDA SDK, if empty the environment variable CUDA_PATH will be used to deduce this path."),
 })
 
-def _init(module_ctx):
-    registrations = {}
+def _find_modules(module_ctx):
+    root = None
+    our_module = None
     for mod in module_ctx.modules:
-        for toolchain in mod.tags.local_toolchain:
-            if not mod.is_root:
-                fail("Only the root module may override the path for the local cuda toolchain")
-            if toolchain.name in registrations.keys():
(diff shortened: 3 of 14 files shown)


MODULE.bazel
```diff
@@ -8,6 +8,10 @@ bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "platforms", version = "0.0.6")
 
 cuda = use_extension("@rules_cuda//cuda:extensions.bzl", "toolchain")
+cuda.local_toolchain(
+    name = "local_cuda",
+    toolkit_path = "",
```
jsharpe, line 13: If a user doesn't have an install at the default path does this cause a failure if they also override it in their own module file? Should we be loading the extension with dev_dependency = True to avoid this if so?

cloudhan: No, this does not trigger analysis error anymore thanks to #209. The generated `local_cuda` repo is as follows
```
root@c578fa5b9b4e ~/rules_cuda (cloudhan/our-module-local-cuda)# tree bazel-rules_cuda/external/_main\~toolchain\~local_cuda/
bazel-rules_cuda/external/_main~toolchain~local_cuda/
|-- BUILD -> /root/rules_cuda/cuda/runtime/BUILD.local_cuda_disabled
|-- REPO.bazel
|-- WORKSPACE
|-- defs.bzl
`-- toolchain
    |-- BUILD
    |-- clang
    |   `-- BUILD
    `-- disabled
        `-- BUILD -> /root/rules_cuda/cuda/templates/BUILD.local_toolchain_disabled

3 directories, 7 files
root@c578fa5b9b4e ~/rules_cuda (cloudhan/our-module-local-cuda)# cat bazel-rules_cuda/external/_main\~toolchain\~local_cuda/BUILD
load("@bazel_skylib//rules:common_settings.bzl", "bool_setting")

package(
    default_visibility = ["//visibility:public"],
)

bool_setting(
    name = "valid_toolchain_found",
    build_setting_default = False,
)
````

PR #267, opened 2024-08-13
Add support for external_includes

### LINKED ISSUE #266: rules_cuda ignores `external_includes` field in compilation context
`external_includes` was added in ~7.0.0 and rules_cuda ignore that field when merging CcInfos, which has the annoying effect that some deps are not made available to the compiler.

This is a placeholder to reference the issue in the code, and maybe propose a fix later.

layus:
With this PR I get a strange error about forbidden internal API.

Fixes #266

### DIFF
--- a/cuda/private/actions/compile.bzl
+++ b/cuda/private/actions/compile.bzl
@@ -76,6 +76,7 @@ def compile(
             include_paths = common.includes,
             quote_include_paths = common.quote_includes,
             system_include_paths = common.system_includes,
+            external_include_paths = common.external_includes,
             defines = common.local_defines + common.defines,
             host_defines = common.host_local_defines + common.host_defines,
             ptxas_flags = common.ptxas_flags,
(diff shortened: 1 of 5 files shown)


cloudhan:
Seems comming from here https://togithub.com/bazelbuild/bazel/blob/3324f06cbd55ffadd497ca21e94ddb30a7675898/src/main/starlark/builtins_bzl/common/cc/cc_common.bzl#L454

cloudhan:
There is a chain of something like `external_includes != _UNBOUND or \`, and the `external_includes` triggers the forbidden internal API..., looking into the `_PRIVATE_STARLARKIFICATION_ALLOWLIST` and you will notice the they are doing really dumb things and violate the open–closed principle by hardcoding the rule name... 🤦

layus:
So it seems that external_includes are part of the internal API, and we are not supposed to use them. Maybe there is a way to merge compilation contexts without tearing it apart and recreating it like we do here ?

For now, I found a simple hotfix:

```bazel
cuda_library(
    ...
    features = ["-external_include_paths"],
)
```

cloudhan:
I think it is workaroundable, you might just want to consume the external_includes (direct on the target or from deps), but dont propagate it forward by the cc_info.

PR #268, opened 2024-08-14
Clarify SRI checksum needed by archive_override()

zifeitong:
Instead of hexadecimal string of checksum, The SRI value must be in the format like `sha256-{base64(checksum)}`. [mdn.io/sri#tools_for_generating_sri_hashes](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#tools_for_generating_sri_hashes) shows a few ways to generate the SRI values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ bazel_dep(name = "rules_cuda", version = "0.2.1")
 # pick a specific version (this is optional an can be skipped)
 archive_override(
     module_name = "rules_cuda",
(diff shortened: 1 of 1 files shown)


jsharpe:
I actually wonder if https://www.w3.org/TR/SRI/ is the more canonical URL for subresource integrity; the bazel docs don't link anywhere when using this term and assume that you know that it is the same format used for web dev.

PR #273, opened 2024-09-06
fix(bcr): add basic presubmit yaml

wep21:
- add basic presubmit yaml to pass bcr ci for the PR generated automatically
- related to https://github.com/bazelbuild/bazel-central-registry/pull/2731


cloudhan:
@wep21 Thanks for the contribution. Based on the BRC PR I think we can just assume it works and a new tag will pass thoes CI and merge as is?

jsharpe (approved):
Yes this should automatically flow through to the BCR next time we release. Thanks for also updating the source repo as well as the BCR!

PR #313, opened 2025-01-20
fix: search for the `sysroot` in compile args

lukasoyen:
Wit the new `cc_rules` based toolchains, the `sysroot` on the toolchain is not set. But we can infer it from the command line args a normal compilation with that toolchain would get.

### DIFF
--- a/cuda/private/cuda_helper.bzl
+++ b/cuda/private/cuda_helper.bzl
@@ -4,6 +4,7 @@ load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_skylib//lib:types.bzl", "types")
 load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
+load("@bazel_tools//tools/build_defs/cc:action_names.bzl", CC_ACTION_NAMES = "ACTION_NAMES")
 load("//cuda/private:action_names.bzl", "ACTION_NAMES")
 load("//cuda/private:artifact_categories.bzl", "ARTIFACT_CATEGORIES")
 load("//cuda/private:providers.bzl", "ArchSpecInfo", "CudaArchsInfo", "CudaInfo", "Stage2ArchInfo", "cuda_archs")
@@ -230,6 +231,36 @@ def _create_common_info(
         transitive_linking_contexts = transitive_linking_contexts,
     )
 
+def _get_sysroot(ctx):
+    cc_toolchain = find_cpp_toolchain(ctx)
+
+    if cc_toolchain.sysroot:
+        return "--sysroot=%s" % cc_toolchain.sysroot
+
+    feature_configuration = cc_common.configure_features(
+        ctx = ctx,
+        cc_toolchain = cc_toolchain,
+        requested_features = ctx.features,
+        unsupported_features = ctx.disabled_features,
+    )
+
+    variables = cc_common.create_compile_variables(
+        cc_toolchain = cc_toolchain,
+        feature_configuration = feature_configuration,
+    )
+
+    cc_flags = cc_common.get_memory_inefficient_command_line(
+        feature_configuration = feature_configuration,
+        action_name = CC_ACTION_NAMES.cpp_compile,
+        variables = variables,
+    )
+
+    for flag in cc_flags:
+        if flag.startswith("--sysroot="):
+            return flag.removeprefix("--sysroot=")
+
+    return None
+
 def _create_common(ctx):
     """Helper to gather and process various information from `ctx` object to ease the parameter passing for internal macros.
 
@@ -243,7 +274,7 @@ def _create_common(ctx):
 
     merged_cc_info = cc_common.merge_cc_infos(cc_infos = [dep[CcInfo] for dep in all_cc_deps])
 
-    cc_toolchain = find_cpp_toolchain(ctx)
+    sysroot = _get_sysroot(ctx)
 
     # gather include info
     includes = merged_cc_info.compilation_context.includes.to_list()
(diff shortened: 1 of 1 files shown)


cuda/private/cuda_helper.bzl
```diff
@@ -243,7 +274,7 @@ def _create_common(ctx):
 
     merged_cc_info = cc_common.merge_cc_infos(cc_infos = [dep[CcInfo] for dep in all_cc_deps])
 
-    cc_toolchain = find_cpp_toolchain(ctx)
+    sysroot = _get_sysroot(ctx)
```
lukasoyen, line 277: I am quite sure that this is the wrong place and should be moved into `_impl` of `cuda_toolchain_config`. Currently this `get_memory_inefficient_command_line()` is called for every? `cuda_compile()` action.

Moving that would require plumbing through a new `sysroot` attribute and possibly the `sysroot` feature in `cuda_toolchain_config` has never been tested. So I hope I can leave that improvement for someone else.

cloudhan:
@lukasoyen Do you know anyway to query the info from cc_toolchain directly. This creates an unused args on each action run. And is quite wasteful... Better way should fuse it into the toolchain configuration.

cloudhan:
Seems to be designed to be accessible from `cc_toolchain` directly

https://github.com/bazelbuild/rules_cc/blob/6be85c266b1df3a5bd38290c814d83ee643185dd/cc/private/rules_impl/cc_flags_supplier_lib.bzl#L57-L62

Then we can attach it to our cuda_info then add a corresponding feature for it.

lukasoyen:
> Seems to be designed to be accessible from `cc_toolchain` directly

I read through that code while figuring things out. I think that is different and not meant for the rule based toolchains.

- [This](https://github.com/bazelbuild/rules_cc/blob/6be85c266b1df3a5bd38290c814d83ee643185dd/cc/toolchains/cc_flags_supplier.bzl#L28) is the corresponding code for the rule based toolchains. It (and the old variant) use `CC_FLAGS_MAKE_VARIABLE_ACTION_NAME`
- `CC_FLAGS_MAKE_VARIABLE_ACTION_NAME` is not used in [`cc_sysroot`](https://github.com/bazelbuild/rules_cc/blob/6be85c266b1df3a5bd38290c814d83ee643185dd/cc/toolchains/args/sysroot.bzl#L20)
- [`cc_toolchain_config`](https://github.com/bazelbuild/rules_cc/blob/6be85c266b1df3a5bd38290c814d83ee643185dd/cc/toolchains/impl/toolchain_config.bzl#L50) does not set `builtin_sysroot`

I think the rule based toolchains are designed in a way to make the `builtin_sysroot` obsolete.

lukasoyen:
> Do you know anyway to query the info from cc_toolchain directly.

As said in https://github.com/bazel-contrib/rules_cuda/pull/313#discussion_r1921941074 I agree this is wasteful and slow. I don't see a reason to put the detection into the toolchain config. The CC toolchain is an attribute already at the detection already queries the CC toolchain for the compiler binary.

I am just not sure how all that plumbing would turn out. But I am up to try if you prefer it that way.

cloudhan:
Sorry for the delay. Was busy working on internal projects.

After examining the cc rules implementation https://github.com/bazelbuild/bazel/blob/a3abc625c78439a5ebf1bb09491627c802f2453d/src/main/starlark/builtins_bzl/common/cc/cc_toolchain_provider_helper.bzl#L198-L204 I think `cc_toolchain.sysroot` is the de facto sysroot what rules_cuda actions can rely on. 

We can just ignore `libc_top_label` things here. See https://github.com/bazelbuild/bazel/blob/a3abc625c78439a5ebf1bb09491627c802f2453d/src/main/java/com/google/devtools/build/lib/rules/cpp/CppConfiguration.java#L796-L800 and https://bazel.build/reference/command-line-reference#flag--grte_top

Then we have a better way to implement this. I will provide you an improved PR for this issue.

cuda/private/cuda_helper.bzl
lukasoyen: rebased on https://github.com/bazel-contrib/rules_cuda/pull/315

keith:
note that the linked PR does not solve the original issue here. The core issue is that the rules based toolchain just doesn't set `sysroot` on the toolchain config ever https://github.com/bazelbuild/rules_cc/blob/d74915024017250e46d95e91a3defc34174effe0/cc/toolchains/impl/toolchain_config.bzl#L69-L89 and instead it passes a flag https://github.com/bazelbuild/rules_cc/blob/91bef569ae45917be8175d4629a6157ff120d494/cc/toolchains/args/sysroot.bzl#L29-L43 but otherwise doesn't treat it as special at all. The alternative to this change would be allowing the user to provide the sysroot another way, theoretically another --@rules_cuda flag would work

keith:
Here's an upstream issue summarizing the state of things https://github.com/bazelbuild/rules_cc/issues/373

but without this change, if you're using a hermetic sysroot, it is not correctly picked up

cloudhan:
I think a `--@rules_cuda` flag might be the prefered way to move forward.

The current implementation is brutal search, it affect every action invocation.
It match a `"--sysroot=<arg>"` with `"--sysroot"` prefix but it might be implemented as `"--sysroot", "<arg>"` in cc rule :(

cloudhan:
@lukasoyen  another (much) better option would be adding a special rule `cuda_cc_sysroot` to figure out the runtime sysroot and then add the target to the generated @cuda repo and then add it as the implicit deps to all cuda rules, this rules provides `CudaInfo`  https://github.com/bazel-contrib/rules_cuda/blob/2e4e33441d24f87f2236158d104574d76381b153/cuda/private/providers.bzl#L64

We somehow need to extend `CudaInfo` a little bit tho.

This will be a little bit complex but it is the ultimate solution.

Issue #265, opened 2024-08-08
[BUG] linking failed using custom cc_toolchain with platform_cpu, cannot detect multiplatform cuda library, 


ZhenshengLee:
### brief

**NOTE**: in the default platform, which is x86_64(k8) toolchain , the compile and linking works.
I wonder if it's a bug or just a misconfiguration during usage of this repo?

### environment

bazel: version7.0.2
cctoolchain: //bazel/toolchains/v5l (a custom cc toolchain for cross compile in aarch64, like https://github.com/f0rmiga/gcc-toolchain/blob/main/toolchain/cc_toolchain_config.bzl)

```sh
├── toolchains
│   └── v5l
│       ├── BUILD
│       ├── v5l.BUILD
│       └── v5l_cc_toolchain_config.bzl
```

### repro steps

simply compile the basic example with cu_library and report the following errors.
**NOTE**: in the default platform, which is x86_64(k8) toolchain , the compile works.

```bazel
cc_binary(
    name = "module_cuda_main",
    srcs = ["tool/module_cuda_main.cpp"],
    includes = ["include"],
    tags = ["tool"],
    visibility = ["//main:__pkg__"],
    deps = [
        ":module_cuda"
    ]
```

```sh
(03:11:14) INFO: Current date is 2024-08-08
(03:11:14) INFO: Analyzed 323 targets (0 packages loaded, 21 targets configured).
(03:11:14) ERROR: /gw_demo/modules/team_demo/module_demo/BUILD:57:15: Linking modules/team_demo/module_demo/module_cuda_main failed: (Exit 1): aarch64-buildroot-linux-gnu-gcc failed: error executing CppLink command (from target //modules/team_demo/module_demo:module_cuda_main) 
  (cd /home/zs/.cache/bazel/_bazel_zs/2c098eac6c684e1fabebb74f5f4483bd/execroot/gaos && \
  exec env - \
    LD_LIBRARY_PATH=/usr/local/cuda/lib64:/usr/local/lib:/usr/lib/x86_64-linux-gnu:/opt/rti.com/rti_connext_dds-6.0.1/lib/x64Linux4gcc7.3.0:/opt/ros/humble/opt/rviz_ogre_vendor/lib:/opt/ros/humble/lib/x86_64-linux-gnu:/opt/ros/humble/lib \
[...]

PR #315, opened 2025-01-28
feat: make sysroot a feature

cloudhan:
Possibly supersede #313.

This should be fully address and tested in the future cross compile support.

### DIFF
--- a/cuda/private/actions/compile.bzl
+++ b/cuda/private/actions/compile.bzl
@@ -75,6 +75,7 @@ def compile(
             cuda_toolchain,
             cuda_feature_config,
             common.cuda_archs_info,
+            common.sysroot,
             source_file = src.path,
             output_file = obj_file.path,
             host_compiler = host_compiler,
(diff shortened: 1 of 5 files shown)


cloudhan:
@lukasoyen I think this will work.

lukasoyen:
I do like this and would rebase my PR once this is merged. Unfortunately this alone does not fix my issue.

I am having a hard time coming up with a reproducer. But my problem comes from using a `rules_cc` based toolchain [like this one](https://github.com/bazelbuild/rules_cc/blob/96d7d59f1fc928ac78b08e07f1172e5b24fe5dac/examples/rule_based_toolchain/toolchains/clang/BUILD.bazel#L19-L33). These new toolchains do not set the `builtin_sysroot` attribute.

cloudhan:
I added some flag tests for sysroot. But since it requires us to register a dummy toolchain with sysroot in it, so without a large refactor, I don't we are ready for the merge...

cloudhan:
Due to complexity for test, leave it for future cross compiling.

ZhenshengLee:
> future cross compiling

good to see that helps to resolve #265

PR #330, opened 2025-03-03
refactor: remove use_cpp_toolchain

udaya2899:
This PR tries to use `use_cpp_toolchain` from `@bazel_tools` instead of the copied definition earlier.

PR #331, opened 2025-03-03
feat: automatically detect compiler_executable from registered cc_toolchain

### LINKED ISSUE #324: Using rules_cuda with hermetic clang
I am trying out #283 from cloudhan/hermetic-ctk-3 branch locally. 
We have a hermetic clang+llvm setup following `@bazel-contrib/toolchains_llvm`:

```
bazel_dep(name = "toolchains_llvm", version = "1.2.0")

llvm = use_extension(
    "@toolchains_llvm//toolchain/extensions:llvm.bzl",
    "llvm",
    dev_dependency = True,
)
llvm.toolchain(
    llvm_version = "18.1.8",
)

use_repo(llvm, "llvm_toolchain")

register_toolchains(
    "@llvm_toolchain//:all",
    dev_dependency = True,
)
```

This is not detected by the `detect_clang` method as it tries to settle for `/usr/bin/clang` and then fails at:
```
Invalid Argument: docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/usr/bin/clang": stat /usr/bin/clang: no such file or directory: unknown.
  Unrecognized error detail: type_url: "type.googleapis.com/google.devtools.remotebuildbot.CommandStatus"
```

How to let `detect_clang` know about this? The clang executable is specifically at:
`@llvm_toolchain//:bin/clang-cpp`. Maybe something like `execpath` could help?

udaya2899:
When a compatible `cc_toolchain` is registered, prefer that when compiler_executable is not set or set to non-hermetic system executable.

### Why?
With this PR's change, I was able to just make my pre-configured `toolchains_llvm`-based toolchain be automatically selected. Without this, I had to fully specifiy my own local toolchain with a `compiler_executable` pointing to the llvm_toolchain's clang binary.

After #330 

Fixes #324 

TODO:

- [x] Create `compiler_use_cc_toolchain` flag to set the value to true while generating the BUILD file
- [ ] Add an integration test using this feature

### DIFF
--- a/cuda/private/toolchain.bzl
+++ b/cuda/private/toolchain.bzl
@@ -1,4 +1,4 @@
-load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
+load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain", "use_cpp_toolchain")
 load("//cuda/private:providers.bzl", "CudaToolchainConfigInfo", "CudaToolkitInfo")
 load("//cuda/private:toolchain_config_lib.bzl", "config_helper")
 
@@ -70,6 +70,7 @@ def _cuda_toolchain_impl(ctx):
 cuda_toolchain = rule(
     doc = """This rule consumes a `CudaToolchainConfigInfo` and provides a `platform_common.ToolchainInfo`, a.k.a, the CUDA Toolchain.""",
     implementation = _cuda_toolchain_impl,
+    toolchains = use_cpp_toolchain(),
     attrs = {
         "toolchain_config": attr.label(
             mandatory = True,
@@ -80,7 +81,7 @@ cuda_toolchain = rule(
         "compiler_executable": attr.string(doc = "The path of the main executable of this toolchain. Either compiler_executable or compiler_label must be specified if compiler_use_cc_toolchain is not set."),
         "compiler_label": attr.label(allow_single_file = True, executable = True, cfg = "exec", doc = "The label of the main executable of this toolchain. Either compiler_executable or compiler_label must be specified."),
         "compiler_files": attr.label(allow_files = True, cfg = "exec", doc = "The set of files that are needed when compiling using this toolchain."),
-        "_cc_toolchain": attr.label(default = "@bazel_tools//tools/cpp:current_cc_toolchain"),
+        "_cc_toolchain": attr.label(default = Label("@bazel_tools//tools/cpp:current_cc_toolchain")),
     },
 )
 
--- a/cuda/private/toolchain.bzl
+++ b/cuda/private/toolchain.bzl
@@ -1,4 +1,4 @@
-load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain", "use_cpp_toolchain")
+load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 load("//cuda/private:providers.bzl", "CudaToolchainConfigInfo", "CudaToolkitInfo")
 load("//cuda/private:toolchain_config_lib.bzl", "config_helper")
 
@@ -21,7 +21,8 @@ def _cuda_toolchain_impl(ctx):
             compiler_executable = cc_toolchain.compiler_executable
         else:
             fail("compiler_use_cc_toolchain set to True but cannot find a configured cc_toolchain")
-    # Next, check for compiler_executable or compiler_label
+
+        # Next, check for compiler_executable or compiler_label
     elif has_compiler_executable:
         compiler_executable = ctx.attr.compiler_executable
     elif has_compiler_label:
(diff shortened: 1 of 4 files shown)


cuda/private/toolchain.bzl
```diff
@@ -1,18 +1,24 @@
...
+        fail("Either compiler_executable or compiler_label must be specified or a valid cc_toolchain must be registered.")
 
     if has_compiler_executable:
         compiler_executable = ctx.attr.compiler_executable
-    else:
+    elif has_compiler_label:
         l = ctx.attr.compiler_label.label
         compiler_executable = "{}/{}/{}".format(l.workspace_root, l.package, l.name)
 
+    if (ctx.attr.compiler_executable == "cuda-clang-not-found" or ctx.attr.compiler_executable.startswith("/")) and has_cc_toolchain:
```
udaya2899, line 19: I also want to cover the case where `which clang` at 
https://github.com/bazel-contrib/rules_cuda/blob/3ace346df43acb3a0e44cc5ed7f5e2cb66cb6c63/cuda/private/repositories.bzl#L219

is detected in the loading phase, and the path is set to `/usr/bin/clang` (or any other system path where clang could be installed) while the execution sandbox environment doesn't actually have that. 

How to effectively handle this without a `startswith("/")`? @cloudhan any suggestions?

cloudhan: How about stop trying to workaround, and just address it in the `config_clang`, make it
```starlark
def detect_clang(repository_ctx): ...  # returns clang_path_or_label

def config_clang(repository_ctx, cuda, clang_path_or_label): ...
```
reuse `CUDA_CLANG_PATH` to accept label should be OK. or just add `CUDA_CLANG_LABEL`

udaya2899:
Thanks for the review earlier. I'll get back to it this week again.

udaya2899:
Sorry for the inactivity, I'm back at this again :)

cuda/private/toolchain.bzl
```diff
@@ -1,18 +1,24 @@
+load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain", "use_cpp_toolchain")
 load("//cuda/private:providers.bzl", "CudaToolchainConfigInfo", "CudaToolkitInfo")
 load("//cuda/private:toolchain_config_lib.bzl", "config_helper")
 
 def _cuda_toolchain_impl(ctx):
+    cc_toolchain = find_cpp_toolchain(ctx)
+    has_cc_toolchain = cc_toolchain != None
```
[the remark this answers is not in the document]
udaya2899: Done. That's better. I introduced a new attr for this.

cuda/private/toolchain.bzl
udaya2899: Okay, I modified both the functions. Please check now.

udaya2899:
I will check the CI and fix later, please let me know if you have concerns over the overall direction of the changes.

cloudhan:
LGTM, but please make all CI green.

udaya2899:
@cloudhan CI passes but I had to make some changes in the toolchain.bzl file to make it work. I'm still yet to test this with our usage of rules_cuda.

Shall we add a test for this so we don't regress? Can you help me suggest what specific test is better for this case?

udaya2899:
Hermetic Clang still doesn't work with the state of this PR, I have some more [forked changes](https://github.com/bazel-contrib/rules_cuda/compare/main...udaya2899:rules_cuda:udaya2899/clang-cuda-path), for which I'll create a new PR for after this is merged. We can create an integration test for testing the overall Clang support.

Currently, this is only one step further that's more like detecting existing toolchain automatically and needs more changes.

udaya2899:
I'm still not super happy with the fact that the only way to use the cc_toolchain from the already configured toolchain is to create a new custom toolchain config. Can we create a flag that sets `compiler_use_cc_toolchain = True`? For example:

`@rules_cuda//:compiler_use_cc_toolchain=true`?

cloudhan:
> @udaya2899 Can we create a flag that sets `compiler_use_cc_toolchain = True`?

Why not? just go for it.

cloudhan:
It is OK to leave the clang tests along for now.

udaya2899:
I tested against our repo's setup and this works. Currently one needs to add this line to the .bazelrc:

`build --repo_env=CUDA_COMPILER_USE_CC_TOOLCHAIN=true` as repo rules cannot fetch flags and env vars are the only way to communicate with them.

I didn't find a good place to add documentation for this, otherwise, I think we're ready :)

udaya2899:
Some tests don't run because Ubuntu 20.04 is retired on Github Action runners. Others seem to pass.

udaya2899:
@freeform-andre sounds good. We'd also prefer that as well. Can you create an issue for that? (Or I can do it too)

freeform-andre:
> @freeform-andre sounds good. We'd also prefer that as well. Can you create an issue for that? (Or I can do it too)

I'm happy to write something up but what aspect specifically? For incompatibility flags I think those should just be added as needed. For the compatibility docs? Yeah, I can start the convo but it will require maintainers to drive it.

cloudhan (approved):
LTGM, thanks for the contribution.

PR #336, opened 2025-03-11
feat: generate version.json to <component_name>/version.json

cloudhan:


udaya2899:
@cloudhan clang requires also a root `version.json` file, not just for individual components. I don't know if clang expects the root version.json to be full-fledged with the version of individual libs or just the main cuda version.

Eitherways, we need the root `version.json` or else clang treats as an invalid `cuda-path`

PR #342, opened 2025-03-18
Add CcInfo external_includes

keith:
These are set if your `cuda_library` depends on a `cc_library` that is
part of another bazel module besides the main one and you set
`--features=external_include_paths`

We could thread these through with a different variable but they end up
being treated the same so I don't think it matters.

### DIFF
--- a/cuda/private/cuda_helper.bzl
+++ b/cuda/private/cuda_helper.bzl
@@ -256,6 +256,7 @@ def _create_common(ctx):
     for inc in attr.includes:
         system_includes.extend(_resolve_includes(ctx, inc))
     system_includes.extend(merged_cc_info.compilation_context.system_includes.to_list())
+    system_includes.extend(merged_cc_info.compilation_context.external_includes.to_list())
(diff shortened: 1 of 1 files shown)


cuda/private/cuda_helper.bzl
```diff
@@ -256,6 +256,7 @@ def _create_common(ctx):
     for inc in attr.includes:
         system_includes.extend(_resolve_includes(ctx, inc))
     system_includes.extend(merged_cc_info.compilation_context.system_includes.to_list())
+    system_includes.extend(merged_cc_info.compilation_context.external_includes.to_list())
```
cloudhan, line 259:
```suggestion
    system_includes.extend(getattr(merged_cc_info.compilation_context, "external_includes", depset()).to_list())
```

cloudhan: Seems some old releases does not have this attr

keith: thanks!

PR #355, opened 2025-04-17
Expand make variables in copts and host_copts

jhance:
This mirrors behavior of rules_cc enabling things like `-I$(GENDIR)/external/foo` for when you need to reference a generated header.

### DIFF
--- a/cuda/private/cuda_helper.bzl
+++ b/cuda/private/cuda_helper.bzl
@@ -276,13 +276,13 @@ def _create_common(ctx):
     # gather compile info
     defines = []
     local_defines = [i for i in attr.local_defines]
-    compile_flags = attr._default_cuda_copts[BuildSettingInfo].value + [o for o in attr.copts if _check_opts(o)]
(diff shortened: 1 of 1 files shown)


cloudhan:
Could you please also add a testcase? See https://github.com/bazel-contrib/rules_cuda/blob/main/tests/flag/BUILD.bazel

jhance:
Sure, but the nature of this means that I would have to hardcode an expansion that is dependent on the configuration you are running the test, right?

cloudhan:
> hardcode an expansion

Yes. You can control the TARGET_CPU and COMPILATION_MODE to construct a test case. see https://github.com/bazel-contrib/rules_cuda/blob/main/tests/flag/flag_validation_test.bzl on how to control them.

Also, `expand_make_variables` is marked as deprecated, it there any modern alternative?

jhance:
Upstream deprecated it without replacement, see https://github.com/bazelbuild/bazel/issues/5859. I am confused as to why it is deprecated.

cloudhan:
Can you fix the CI failures, seems everyone is relying on the deprecated feature, as long as it is not removed.

PR #356, opened 2025-04-17
Update deprecated CI runners

freeform-andre:
Account for deprecation of `ubuntu-20.04`
https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/

### DIFF
--- a/.github/workflows/build-tests.yaml
+++ b/.github/workflows/build-tests.yaml
@@ -14,10 +14,10 @@ jobs:
     strategy:
       matrix:
         cases:
-          - { os: "ubuntu-20.04", cuda-version: "11.2.2", source: "nvidia" }
+          - { os: "ubuntu-22.04", cuda-version: "11.2.2", source: "nvidia" }
           - { os: "ubuntu-22.04", cuda-version: "11.8.0", source: "nvidia" }
           - {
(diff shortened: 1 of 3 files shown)


freeform-andre:
```
/usr/bin/sudo apt-get -y install cuda-nvcc-11-6 cuda-cudart-dev-11-6
Reading package lists...
Building dependency tree...
Reading state information...
Package cuda-nvcc-11-6 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'cuda-nvcc-11-6' has no installation candidate
E: Unable to locate package cuda-cudart-dev-11-6
```
Does this suggest this version is now not testable on github runners?

freeform-andre:
cc @cloudhan for context

cloudhan:
Yeah, 11.6 is not available. See https://developer.download.nvidia.cn/compute/cuda/repos/ubuntu2204/x86_64/

Maybe just bump to 11.7, 11.6 was select due to some new flags was added in that release. But it is not important nowadays.

freeform-andre:
@cloudhan I think I updated the right things.

Issue #358, opened 2025-04-25
Enable buildifier


UebelAndre:
Any objections to enabling [buildifier](https://github.com/bazelbuild/buildtools/blob/v8.0.3/buildifier/README.md) and addressing all lints? Such a change would be likely be large but the changes would be pretty well defined guided by the rules of the lints.

cloudhan:
No, this project basically has only one or two active developers. Formatting is already enforced in pre-commit workflow. For linting, I am highly against it due to past projects experience. Conforming to all of those various linting 'standards' has not increased the speed of development, nor the robustness, nor the client satisfaction.

UebelAndre:
I would agree all lints are not all that useful but surely there are ones such as `uninitialized` or `unused-variable` which in general are bugs waiting to be found. Buildifier supports a config file where you can specify the lints you care about, would having that as a curated list be a happy middle ground?

cloudhan:
I don't have time at the moment, but it is welcomed to add those linting rules that actually fix things :)

UebelAndre:
I can do that. Also FYI, I’m pinging issues because I’m interested in contributing but I need some guidance. Thank you for all the work you’ve put into the project thus far and I’m going I can help 😄

PR #359, opened 2025-04-27
Addressed all buildifier warnings

UebelAndre:
This change applies buildifier auto-fixes for all available fixes and then in a second commit manually updates implementations to conform to the warning.

Additionally `.buildifier.json` has been added with the outputs of `buildifier -config=example`. Maintainers should remove anything there that offend them.

closes https://github.com/bazel-contrib/rules_cuda/issues/358


cloudhan:
please disable all those stylistic linting and revert all those style change.

UebelAndre:
If you can be specific about the warnings I can remove them but if this is still a controversial change I can close it out.

cloudhan:
Is it possible to make it opt-in first then we gradually enable more of them?

PR #370, opened 2025-06-24
Remove duplicate cupti_header target

lgulich:


lgulich:
Hi @cloudhan 

I've noticed that the target `cupti_headers` is currently duplicated. It is generated once by the cupti template [here](https://github.com/bazel-contrib/rules_cuda/blob/9e65ec3da00698e15fb3b2f31650c1f39438ef6e/cuda/private/templates/BUILD.cupti#L14) and once by the cuda headers template [here](https://github.com/bazel-contrib/rules_cuda/blob/9e65ec3da00698e15fb3b2f31650c1f39438ef6e/cuda/private/templates/BUILD.cuda_headers#L10)

Consequently bazel throws an error that there are conflicting targets. Do you know why cupti requires a separate header target? Is it safe to remove it?

Thanks

cuda/private/templates/BUILD.cupti
```diff
@@ -11,15 +11,6 @@ cc_import(
     target_compatible_with = ["@platforms//os:windows"],
 )
 
-cc_library(
-    name = "cupti_headers",
```
cloudhan, line 15: Please just rename it to `cupti_headers`

cloudhan: And the `if_windows([` part in `cupti` target

lgulich: thanks, done

PR #360, opened 2025-04-29
feat: hermetic CUDA clang support 

udaya2899:
Partially part of #283 

This PR adds the missing piece of hermetic CUDA compilation with clang. The provided `compiler_deps` repo (alias cuda_nvcc) wasn't enough for clang to infer everything from. Typically, it's documented in multiple places ([(1)](https://llvm.org/docs/CompileCudaWithLLVM.html#id3), [(2)](https://reviews.llvm.org/D42642)) that clang expects the entire CUDA toolkit and there is no recommended "minimal" CUDA setup. Nevertheless, I found that `cccl`, `nvcc`, `cudart`, `nvvm` components were enough.

Surprisingly, I didn't need the `version.json` or `version.txt` file anymore so I removed the logic to generate the root `version.json` file.

In `config_clang` we create a new folder called `clang_compiler_deps` where we collect `bin`, `lib`, `include` files from the required components and place in a single collected folder. 

We've been using this implementation for more than a month now with over 20+ `cuda_library` usages and `cc_test`s that run on NVIDIA GPU hermetically in RBE with CUDA 12.3 and LLVM 19.1.0

PR #375, opened 2025-07-13
refactor!: rework hermetic clang

### LINKED ISSUE #337: Using cuda with clang compiler expects curand to be added explicitly
When compiler is clang, curand should be implicitly added somewhere (maybe `@local_cuda//:compiler_deps`?). Currently every `cuda_library` needs to add `@local_cuda_curand`.

cloudhan:
Previously, in #360 a folder structure is restored from distributed components for clang to work with hermetic cuda.

Based on LLVM 12 https://github.com/llvm/llvm-project/blob/fed41342a82f5a3a9201819a82bf7a48313e296b/clang/lib/Driver/ToolChains/Cuda.cpp#L180-L223, it ultimately only requires nvcc component folder structure and `nvvm/libdevice/libdevice.10.bc` file to present to make it a valid cuda toolkit for clang.

Additionly, we expose all required components at build time.

This also fixes #337.

### DIFF
--- a/cuda/private/template_helper.bzl
+++ b/cuda/private/template_helper.bzl
@@ -182,27 +182,29 @@ def _generate_toolchain_clang_build(repository_ctx, cuda, clang_path_or_label):
 
     compiler_files = []
     cuda_path_for_subst = ""
-    cuda_path_data = None
+    path_data = None
     if cuda.path:
         compiler_files.append("@cuda//:compiler_deps")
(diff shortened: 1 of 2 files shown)


cloudhan:
To test:

```
bazel build --repo_env=CC=clang --repo_env=CUDA_COMPILER_USE_CC_TOOLCHAIN=true --cuda_compiler=clang <...>
```

PR #382, opened 2025-07-24
If no components specified, use all components.

finn-ball:
If the user does not specify the components to be downloaded, it can default to all of them.

### DIFF
--- a/cuda/extensions.bzl
+++ b/cuda/extensions.bzl
@@ -37,7 +37,7 @@ cuda_component_tag = tag_class(attrs = {
 
 cuda_redist_json_tag = tag_class(attrs = {
     "name": attr.string(mandatory = True, doc = "Repo name for the cuda_redist_json"),
-    "components": attr.string_list(mandatory = True, doc = "components to be used"),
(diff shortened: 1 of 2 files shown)


cuda/extensions.bzl
```diff
@@ -94,6 +94,9 @@ def _module_tag_to_dict(t):
 
 def _redist_json_impl(module_ctx, attr):
     url, json_object = redist_json_helper.get(module_ctx, attr)
+    # Sets the default to be all the components contained if not specified.
+    if attr.components == None:
+        attr.components = json_object.keys()
```
cloudhan, line 99: keys are not available.

We have builtin mapping for it, see registry.bzl

finn-ball: I've had another go which attempts to match components you specify in the `registry.bzl`

cuda/private/redist_json_helper.bzl
```diff
@@ -71,16 +71,25 @@ def _collect_specs(ctx, attr, redist, the_url):
     """
 
     specs = []
```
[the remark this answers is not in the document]
cloudhan: I think this would be enough. Much cleaner.

PR #390, opened 2025-08-02
feat: search for the sysroot in compile args in toolchain rule

cloudhan:
supersede #313 and #322


cloudhan:
It seems I can fold the `cuda_cc_sysroot` into the `cuda_toolchain` at all, 🤔

Issue #395, opened 2025-09-26
Cross-compilation support


ahans:
First up, thanks for this great project, we've been using it successfully for quite a while now!

I'd like to add support for cross compilation for Jetson/aarch64 using `redist.json` as a source. I have hacked already a bit and have a version that can at least compile `examples/basic`. The libraries still need some work, but before I put that in, I'd like to hear if there are already thoughts on how to accomplish that. I guess I could finish my version and it'd be fine for my needs, but ideally I'd like to contribute a PR so that others benefit from this as well.

Here's a rough outline of what I did:

* Extend `redist_json_helper` so that it does not filter for `x86_64`, but also includes `aarch64`. I then added a field `arch` to the struct that gets appended to `specs` for each component.
* Then in `_redist_json_impl` in `cuda/extensions.bzl`, I modify the specs list a bit so that for each component I have an entry where there are dicts `urls`, `shas256`, and `strip_prefixes` that each have two keys, `x86_64` and `aarch64`.
* The result gets fed into a new repository rule `cuda_multi_platform_component` which iterates over the keys, effectively calling `download_and_extract` twice. 
* Finally, the templates needed a bit of work to have targets for each architecture that are used by a top-level target (such as `cuda_runtime`), which has a bunch of selects, very similar to what is currently present to support both Windows and Linux.

As I said, this seems to be working, however, I'm not certain which usecases I broke along the way. Also, some cleanup and refactoring would be necessary to avoid redundancy.

One thing that complicates things a bit is that for most things I need the target architecture archive, but `nvcc` I need for the host. Just like clang, you only need to care about the host architecture here. In the repository rule I extract archives into `x86_64` and `aarch64` subdirectories. For `nvcc` I added some special handling, there I also create some symlinks to the directories in `x86_64`, so that all files are available where current `main` expects them.

Another thing I thought about was having subdirectories only in `lib`, but then I'd have to modify the original archive structure, which is something I'd like to avoid.

Initially I also thought about having a repository per architecture, but then I'd need a new place for the select. Ultimately, I want to be able to depend on `@cuda//:cublas` in my top-level target and it picks the correct libraries automatically depending on the architecture I'm building for.

I only need support for Linux x86_64 as the host and x86_64 and aarch64 as targets. However, AFAIU, support for a Linux aarch64 host would be possible with the redist.json archives, but would complicate the work required for rules_cuda even further.

[...]

PR #396, opened 2025-09-29
Implement initial cross-compilation support

ahans:
Just a very rough draft. For my purposes this might already be good enough, but I'm sure we need to iterate a bit to get this into a mergeable state for this repo. Looking forward to your comments @cloudhan .

#395

### DIFF
--- a/cuda/extensions.bzl
+++ b/cuda/extensions.bzl
@@ -57,6 +57,7 @@ cuda_redist_json_tag = tag_class(attrs = {
         doc = "Generate a URL by using the specified version." +
               "This URL will be tried after all URLs specified in the `urls` attribute.",
     ),
+    "archs": attr.string_list(doc = "Target architectures to support. If not specified, only x86_64 will be supported."),
 })
 
 cuda_toolkit_tag = tag_class(attrs = {
@@ -100,7 +101,8 @@ def _redist_json_impl(module_ctx, attr):
     mapping = {}
     for spec in component_specs:
         repo_name = redist_json_helper.get_repo_name(module_ctx, spec)
-        mapping[spec["component_name"]] = "@" + repo_name
+        component_name = spec["component_name"]
+        mapping["{}__{}".format(component_name, spec["arch"])] = "@" + repo_name
 
         attr = {key: value for key, value in spec.items()}
         attr["name"] = repo_name
--- a/cuda/extensions.bzl
+++ b/cuda/extensions.bzl
@@ -102,7 +102,7 @@ def _redist_json_impl(module_ctx, attr):
     for spec in component_specs:
         repo_name = redist_json_helper.get_repo_name(module_ctx, spec)
         component_name = spec["component_name"]
-        mapping["{}__{}".format(component_name, spec["arch"])] = "@" + repo_name
+        mapping[redist_json_helper.get_repo_mapping_key(component_name, spec["arch"])] = "@" + repo_name
 
         attr = {key: value for key, value in spec.items()}
         attr["name"] = repo_name
--- a/cuda/private/repositories.bzl
+++ b/cuda/private/repositories.bzl
@@ -94,13 +94,15 @@ def _detect_deliverable_cuda_toolkit(repository_ctx):
     required_components = ["cccl", "cudart", "nvcc"]
     for rc in required_components:
         for arch in repository_ctx.attr.archs:
-            if "{}__{}".format(rc, arch) not in repository_ctx.attr.components_mapping:
+            mapping_key = redist_json_helper.get_repo_mapping_key(rc, arch)
+            if mapping_key not in repository_ctx.attr.components_mapping:
                 fail('component "{}" for {} is required.'.format(rc, arch))
 
     nvcc_repo = repository_ctx.attr.components_mapping.get("nvcc", None)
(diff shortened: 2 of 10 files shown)


cuda/private/repositories.bzl
```diff
@@ -276,6 +283,7 @@ cuda_toolkit = repository_rule(
         "nvcc_version": attr.string(
             doc = "nvcc version. Required for deliverable toolkit only. Fallback to version if omitted.",
         ),
+        "archs": attr.string_list(doc = "list of host target architectures to support (e.g., x86_64)"),
```
ahans, line 286: This is not fully plumbed yet.

examples/MODULE.bazel
```diff
@@ -4,21 +4,98 @@ module(
...
+bazel_dep(name = "platforms", version = "0.0.11")
+
 bazel_dep(name = "rules_cuda", version = "0.2.3")
 local_path_override(
     module_name = "rules_cuda",
     path = "..",
 )
 
 cuda = use_extension("@rules_cuda//cuda:extensions.bzl", "toolchain")
+cuda.redist_json(
```
ahans, line 16: I extended the examples to test this. We will probably need something in `tests/integration`?

cuda/private/repositories.bzl
```diff
@@ -90,10 +93,14 @@ def _detect_deliverable_cuda_toolkit(repository_ctx):
...
-            fail('component "{}" is required.'.format(rc))
+        for arch in repository_ctx.attr.archs:
+            if "{}__{}".format(rc, arch) not in repository_ctx.attr.components_mapping:
+                fail('component "{}" for {} is required.'.format(rc, arch))
 
-    nvcc_repo = repository_ctx.attr.components_mapping["nvcc"]
+    nvcc_repo = repository_ctx.attr.components_mapping.get("nvcc", None)
+    if not nvcc_repo:
+        host_arch = "aarch64" if _is_aarch64(repository_ctx) else "x86_64"
+        nvcc_repo = repository_ctx.attr.components_mapping["nvcc__{}".format(host_arch)]
```
cloudhan, line 103: Here we need a bit extension. Previously for non cross compiling case, `@nvcc` are used for both compiling (on **exec** platform) and linking (for **target**  platform). We need to separate them to `@nvcc_the_compiler` `@nvcc_the_library` with better name for sure.

Then your change becomes
```
nvcc_repo = repository_ctx.attr.components_mapping["nvcc_the_compiler"]
```
and your hardcoded `host_arch` goes into a `select` based on `@platforms`

cuda/extensions.bzl
```diff
@@ -100,7 +101,8 @@ def _redist_json_impl(module_ctx, attr):
     mapping = {}
     for spec in component_specs:
         repo_name = redist_json_helper.get_repo_name(module_ctx, spec)
-        mapping[spec["component_name"]] = "@" + repo_name
+        component_name = spec["component_name"]
+        mapping["{}__{}".format(component_name, spec["arch"])] = "@" + repo_name
```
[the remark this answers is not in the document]
ahans: Do what exactly? Return a tuple with repo name and mapping key? Or add a new function `get_mapping_key` (I think that's what I'd prefer).

cuda/private/repositories.bzl
ahans: Yeah, "somehow". 😉 I checked [toolchains_llvm](https://github.com/bazel-contrib/toolchains_llvm). They also have to select the correct distribution depending on the host, looks like this happens in [repository rules implementation](https://github.com/bazel-contrib/toolchains_llvm/blob/master/toolchain/internal/llvm_distributions.bzl) based on `repository_ctx`. There could be a way to override this (host vs exec platform), but it doesn't look like a `select` is part of any of this.

ahans: In `MODULE.bazel`, we need this:

```
host_platform = use_extension("@platforms//host:extension.bzl", "host_platform")
use_repo(host_platform, "host_platform")
```

Then in some `.bzl` or `BUILD` file we can do this:
```
load("@host_platform//:constraints.bzl", "HOST_CONSTRAINTS")
```

On my machine, `constraints.bzl` looks like this:
```

# DO NOT EDIT: automatically generated constraints list
HOST_CONSTRAINTS = [
  '@platforms//cpu:x86_64',
  '@platforms//os:linux',
]
```

If we could put that on the left side of a `select` in some `nvcc` meta-target, we´d be done. But that left side gets evaluated against the target platform, so really not sure how that's supposed to work. I guess we could load this in a `.bzl` file and then manually implement the `select` in the repository rule. Would probably be cleaner than using `repository_ctx` to get that info, but the logic afterwards would remain the same.

ahans: Could it be that I don't need any of this, but instead need to extend the `BUILD.toolchain_nvcc` template (and code that uses it) to register both, `nvcc_x86_64` and `nvcc_aarch64`, as separate toolchains, but with different `exec_compatible_with` entries (one with `@platforms//cpu:x86_64`, the other with `@platforms//cpu:aarch64`). And then Bazel would pick the right one automatically depending on the exec platform?

ahans:
Superseded by #454

PR #416, opened 2025-11-20
Include user link flags in device links

JamesMBartlett:
I was seeing undefined reference errors when trying to device link a cuda file with NVSHMEM, eg.:
```
nvlink error   : Undefined reference to 'nvshmemi_device_state_d'
```

This PR adds user link flags to the device link command line, so that eg. `-lnvshmem` can be passed down to the device linking, and it fixes the error I encountered above.


cloudhan:
Sorry for the delay. Would you mind porting it to nvcc_msvc and clang config?

PR #419, opened 2025-12-02
Retain cc toolchain flags

### LINKED ISSUE #354: Hermetic toolchain seems to pick up cc compiler executable but not flags
This probably works in a lot of default cases, but not generally. For example, our compiler ([portable_cc_toolchain](https://registry.bazel.build/modules/portable_cc_toolchain)) uses a sysroot that requires the flags `--sysroot=path/to/sysroot --gcc-toolchain=path/to/gcc` among a few other required linker flags. I think the correct implementation would use something like [`get_memory_inefficient_command_line`](https://bazel.build/rules/lib/toplevel/cc_common#get_memory_inefficient_command_line) for both compile and link actions, and add those flags via `-Xcompiler` and `-Xlinker`.

calebzulawski:
This is necessary with my cc toolchain to retain various flags configured by the toolchain that aren't possible to pass via copts.  My toolchain requires `--gcc-toolchain` in addition to `--sysroot` which prevents us from using rules_cuda. This change also has the added benefit of inheriting important toolchain flags like reproducibility flags and supporting cc toolchain features like sanitizers.

Closes #354

### DIFF
--- a/cuda/private/cuda_helper.bzl
+++ b/cuda/private/cuda_helper.bzl
@@ -260,6 +260,8 @@ def _get_cc_host_compile_flags(ctx):
     dummy_out = "__rules_cuda_host_compile_output__"
     user_flags = ctx.fragments.cpp.copts + ctx.fragments.cpp.cxxopts
 
+    cuda_toolchain = find_cuda_toolchain(ctx)
+
     cc_toolchain = find_cpp_toolchain(ctx)
     if cc_toolchain == None:
         return []
@@ -285,20 +287,39 @@ def _get_cc_host_compile_flags(ctx):
         variables = variables,
     )
 
+    flags_to_skip = [
+        "-c",
+    ]
+
+    if cuda_toolchain.toolchain_identifier == "clang":
+        flags_to_skip.append(
+            "-fno-canonical-system-headers",
+        )
+
+    flag_prefixes_to_skip = [
+        "-std=",
+        "--sysroot=",
+    ]
+
+    flags_with_value_to_skip = [
+        "--sysroot",
+        "-o",
+    ]
+
     filtered_flags = []
     skip_next = False
     for flag in command_line:
         if skip_next:
             skip_next = False
             continue
-        if flag == "-c":
+        if flag in flags_to_skip:
+            continue
+        if flag in flags_with_value_to_skip:
+            skip_next = True
             continue
         if dummy_src in flag or dummy_out in flag:
             continue
-        if flag in ["--sysroot", "-o"]:
-            skip_next = True
-            continue
-        if any([flag.startswith(f) for f in ("-std=", "--sysroot=")]):
+        if any([flag.startswith(f) for f in flag_prefixes_to_skip]):
             continue
         filtered_flags.append(flag)
 
--- a/cuda/private/toolchain.bzl
+++ b/cuda/private/toolchain.bzl
@@ -112,6 +112,7 @@ def _cuda_toolchain_impl(ctx):
             selectables_info = selectables_info,
             artifact_name_patterns = artifact_name_patterns,
             cuda_toolkit = cuda_toolchain_config.cuda_toolkit,
+            toolchain_identifier = cuda_toolchain_config.toolchain_identifier,
             **optional_attributes
         ),
     ]


calebzulawski:
That should fix the error. `-std` shouldn't be passed via `-Xcompiler` (but might deserve its own feature in the future similar to sysroot). Seems Bazel 7 adds that flag but 8 doesn't.

cuda/private/cuda_helper.bzl
```diff
@@ -250,6 +255,55 @@ def _get_cc_sysroot(ctx):
     cuda_toolchain = find_cuda_toolchain(ctx)
     return getattr(cuda_toolchain, "cc_sysroot", None)
 
+def _get_cc_host_compile_flags(ctx):
```
cloudhan: I believe you want to move this to toolchain.bzl, together with `_get_cc_sysroot`, otherwise, you will need to query those flags on every cuda related actions!

calebzulawski: I think we do need to do that, since the flags are dependent on the features enabled for the particular rule

calebzulawski:
> Could you please list a very concrete cmdline as an example why we need this for potential future reader.

Using my toolchain (https://registry.bazel.build/modules/portable_cc_toolchain) `examples/basic/kernel.cu` with a hermetic nvcc produces the following commandline:

```bash
external/rules_cuda++toolchain+cuda_nvcc_v13.0.48/nvcc/bin/nvcc -x cu -ccbin external/portable_cc_toolchain++assets+llvm-x86_64-unknown-linux-gnu/bin/clang++ -I . -I bazel-out/k8-fastbuild/bin -I external/rules_cuda++toolchain+cuda_cudart_v13.0.48 -I bazel-out/k8-fastbuild/bin/external/rules_cuda++toolchain+cuda_cudart_v13.0.48 -I external/rules_cuda++toolchain+cuda_nvcc_v13.0.48 -I bazel-out/k8-fastbuild/bin/external/rules_cuda++toolchain+cuda_nvcc_v13.0.48 -I external/rules_cuda++toolchain+cuda_cccl_v13.0.50 -I bazel-out/k8-fastbuild/bin/external/rules_cuda++toolchain+cuda_cccl_v13.0.50 -I external/rules_cuda++toolchain+cuda_crt_v13.0.48 -I bazel-out/k8-fastbuild/bin/external/rules_cuda++toolchain+cuda_crt_v13.0.48 -I external/rules_cuda++toolchain+cuda_culibos_v13.0.39 -I bazel-out/k8-fastbuild/bin/external/rules_cuda++toolchain+cuda_culibos_v13.0.39 -isystem external/rules_cuda++toolchain+cuda_nvcc_v13.0.48/nvcc/include -isystem bazel-out/k8-fastbuild/bin/external/rules_cuda++toolchain+cuda_nvcc_v13.0.48/nvcc/include -isystem external/rules_cuda++toolchain+cuda_cccl_v13.0.50/cccl/include -isystem bazel-out/k8-fastbuild/bin/external/rules_cuda++toolchain+cuda_cccl_v13.0.50/cccl/include -isystem external/rules_cuda++toolchain+cuda_crt_v13.0.48/crt/include -isystem bazel-out/k8-fastbuild/bin/external/rules_cuda++toolchain+cuda_crt_v13.0.48/crt/include -isystem external/rules_cuda++toolchain+cuda_cudart_v13.0.48/cudart/include -isystem bazel-out/k8-fastbuild/bin/external/rules_cuda++toolchain+cuda_cudart_v13.0.48/cudart/include -O0 -Xcompiler -g1 -Xcompiler '--target=x86_64-unknown-linux-gnu' -Xcompiler '--gcc-toolchain=external/portable_cc_toolchain++assets+sysroot-x86_64-unknown-linux-gnu' -Xcompiler -no-canonical-prefixes -Xcompiler '-fdiagnostics-color=always' -Xcompiler -Wno-builtin-macro-redefined -Xcompiler '-D__DATE__="redacted"' -Xcompiler '-D__TIMESTAMP__="redacted"' -Xcompiler '-D__TIME__="redacted"' -Xcompiler '-ffile-compilation-dir=.' -Xcompiler '-ffile-prefix-map=external/portable_cc_toolchain++assets+sysroot-x86_64-unknown-linux-gnu=/sysroot' -Xcompiler -O0 -Xcompiler -Wall -Xcompiler -gmlt -Xcompiler '--sysroot=external/portable_cc_toolchain++assets+sysroot-x86_64-unknown-linux-gnu/x86_64-unknown-linux-gnu/sysroot' -c basic/kernel.cu -o bazel-out/k8-fastbuild/bin/basic/_objs/kernel/kernel.o --expt-relaxed-constexpr --expt-extended-lambda
```

The most important flags there are `-Xcompiler '--target=x86_64-unknown-linux-gnu' -Xcompiler '--gcc-toolchain=external/portable_cc_toolchain++assets+sysroot-x86_64-unknown-linux-gnu'` as the target changes based on a config setting (could be musl) and the `--gcc-toolchain` flag needs to be propagated to properly find libstdc++. There are some additional nice-to-have flags that appear like ` '-ffile-prefix-map=external/portable_cc_toolchain++assets+sysroot-x86_64-unknown-linux-gnu=/sysroot'` which would also be impossible to manually add to every rule.  Flags associated with features are also now passed, such as my toolchain's feature to enable asan, or the reproducibility flags `-Xcompiler -Wno-builtin-macro-redefined -Xcompiler '-D__DATE__="redacted"' -Xcompiler '-D__TIMESTAMP__="redacted"' -Xcompiler '-D__TIME__="redacted"' -Xcompiler '-ffile-compilation-dir=.'`.

calebzulawski:
It seems the one remaining failure is due to using gcc as the cc toolchain but clang as the cuda compiler: https://github.com/bazel-contrib/rules_cuda/actions/runs/19882740086/job/56984741862#step:21:69. Clang doesn't have GCC's `-fno-canonical-system-headers` flag.

I think the easiest solution is "don't do this", i.e. use clang as both your cc and cuda compiler, or use nvcc and use your cc compiler to ensure that the flags are consistent.  If a user insists on doing this they can always just disable the `cuda_host_use_toolchain_flags` feature entirely, but maybe we should filter out known non-clang flags when using clang as the cuda compiler? I don't think we can reasonably filter out every flag, but we can at least filter out the defaults added by Bazel.

calebzulawski:
@cloudhan looks like this might be good now. I added a section to filter out gcc-only flags in that particular situation. A user can always disable the feature for host compiler flags entirely, but I think this is a pretty good default behavior.

PR #472, opened 2026-05-02
Auto-detect CUDA platform from @platforms constraints

linzhp:
## What changes

  - **New `--@rules_cuda//cuda:aarch64=sbsa|aarch64` flag** disambiguates `linux-sbsa` (server-class ARM, discrete GPU on PCIe — Grace, A4X, Graviton+GPU) from `linux-aarch64` (NVIDIA's Tegra/Jetson/Drive embedded redist with on-die GPU) when the active platform is aarch64-linux. Default is `sbsa` (linux-sbsa) — the safer fallback since the sbsa redist's libraries don't depend on Tegra-specific kernel modules. The flag is a no-op on x86   
  and Windows platforms, which are unambiguously identified by `@platforms//cpu` + `@platforms//os` alone.
                                                                                                                                              
  - **`target_platform` flag removed.** Fully replaced by `--platforms` + optional `--@rules_cuda//cuda:aarch64`.

After CUDA 13 or above, when Linux sbsa and aarch64 are unified, people no longer need to pass `--@rules_cuda//cuda:aarch64`. 
                                                                                                                                              
  ## Usage                                                                                                                                    
                                               
  ```
  # Vanilla linux-x86_64 — no extra flags needed.
  bazel build --platforms=//:my_x86_platform //:cuda_target                                                                                              
  
  # Server-class ARM (Grace, A4X, Graviton + dGPU).                                                                                                      
  # :aarch64 defaults to sbsa → resolves to linux-sbsa.                                                                                                  
  bazel build --platforms=//:my_aarch64_platform //:cuda_target                                                                                          
                                                                                                                                                         
  # Drive Thor / Jetson Tegra (on-die GPU).                                                                                                              
  bazel build --platforms=//:my_aarch64_platform \                                                                                                       
              --@rules_cuda//cuda:aarch64=aarch64 \                                                                                                      
              //:cuda_target                                                                                                                  
  ```                                                                                                                                                    
                                                                                                                                                         
  ## Breaking change                                                                                                                                     
                                                                                                                                                         
  `--@rules_cuda//cuda:target_platform=<plat>` is removed. Callers should switch to `--platforms` + `--@rules_cuda//cuda:aarch64` where needed:
                                                                                                                                                         
  | Before | After |                           
  |---|---|                                                                                                                                              
  | `--@rules_cuda//cuda:target_platform=linux-x86_64` | `--platforms=//:my_x86_platform` |
  | `--@rules_cuda//cuda:target_platform=linux-sbsa` | `--platforms=//:my_aarch64_platform` |                                                            
  | `--@rules_cuda//cuda:target_platform=linux-aarch64` | `--platforms=//:my_aarch64_platform --@rules_cuda//cuda:aarch64=aarch64` |                     
                                                                                                                                                         
  The flag was only introduced in #454 and hasn't shipped in a tagged release yet, so impact should be limited.                                
                                                                                                                                              
  ## Verified                                                                                                                                            
                                                                                     
  Built downstream against a real cross-compile setup (x86 host → x86 / ARM-server-class / Tegra targets) with all three target platforms resolving to the correct CUDA redist via `cquery`:                                                                                                                  
  
  | `--platforms` + flags | resolved `cuda_runtime` redist |                                                                                             
  |---|---|                                                                                                                                              
  | `//:linux_amd64` | `cuda_cudart_linux_x86_64_12_8_1` |                                                                                    
  | `//:linux_aarch64` (default `aarch64=sbsa`) | `cuda_cudart_linux_sbsa_12_8_1` |                                                                      
  | `//:linux_aarch64` + `aarch64=aarch64` | `cuda_cudart_linux_aarch64_12_8_1` |                                                                        
                                                                                                                                                         
  Per-platform `cuda_nvcc` binaries (CUDA 12.8.93) confirmed native to their declared platform — `linux-x86_64` is x86_64 ELF, `linux-sbsa` and `linux-aarch64` are both aarch64 ELF (different glibc versions, not interchangeable). Confirms the existing `platform_type = "exec"` classification for nvcc/nvvm is correct across all four supported platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,10 @@ use_repo(cuda, "cuda")
 #### Multi-version hermetic toolchains (Bzlmod)
 
 To select CUDA versions and platforms at build time, define multiple `cuda.redist_json` entries and a single `cuda.toolkit`.
-Then use `@rules_cuda//cuda:version`, `@rules_cuda//cuda:exec_platform`, and `@rules_cuda//cuda:target_platform` flags.
-If `@rules_cuda//cuda:version` is unset, rules_cuda selects the highest declared redist version.
+The target CUDA redist is auto-detected from the active target platform's `@platforms//cpu` + `@platforms//os` + `@rules_cuda//cuda:gpu_kind` constraints —
+set `--platforms` to control it (predefined platforms `@rules_cuda//cuda:linux_x86_64`, `:linux_sbsa`, `:linux_aarch64`, `:windows_x86_64` are provided).
+Use `@rules_cuda//cuda:version` to pick the CUDA version (defaults to the highest declared if unset),
+and `@rules_cuda//cuda:exec_platform` as an escape hatch to override the auto-detected exec redist.
 
 ```starlark
 cuda = use_extension("@rules_cuda//cuda:extensions.bzl", "toolchain")
@@ -65,8 +67,7 @@ use_repo(cuda, "cuda")
 Example `.bazelrc` entries:
 
 ```
-build --@rules_cuda//cuda:exec_platform=linux-x86_64
-build --@rules_cuda//cuda:target_platform=linux-x86_64
+build --platforms=@rules_cuda//cuda:linux_x86_64
 # Optional: if omitted, the highest declared redist version is used.
 build --@rules_cuda//cuda:version=13.0.0
 ```
--- a/README.md
+++ b/README.md
@@ -35,8 +35,9 @@ use_repo(cuda, "cuda")
 #### Multi-version hermetic toolchains (Bzlmod)
 
 To select CUDA versions and platforms at build time, define multiple `cuda.redist_json` entries and a single `cuda.toolkit`.
-The target CUDA redist is auto-detected from the active target platform's `@platforms//cpu` + `@platforms//os` + `@rules_cuda//cuda:gpu_kind` constraints —
-set `--platforms` to control it (predefined platforms `@rules_cuda//cuda:linux_x86_64`, `:linux_sbsa`, `:linux_aarch64`, `:windows_x86_64` are provided).
+The target CUDA redist is auto-detected from the active target platform's `@platforms//cpu` + `@platforms//os` constraints.
+Set `--platforms` to control it. linux-sbsa and linux-aarch64 share aarch64+linux constraints; the `@rules_cuda//cuda:target_gpu` flag
+disambiguates them (default `discrete` → linux-sbsa, set to `on_die` for Tegra/Jetson/Drive Thor → linux-aarch64).
 Use `@rules_cuda//cuda:version` to pick the CUDA version (defaults to the highest declared if unset),
 and `@rules_cuda//cuda:exec_platform` as an escape hatch to override the auto-detected exec redist.
 
@@ -67,7 +68,9 @@ use_repo(cuda, "cuda")
 Example `.bazelrc` entries:
 
 ```
-build --platforms=@rules_cuda//cuda:linux_x86_64
+build --platforms=//:my_x86_platform
+# For Tegra/Drive Thor targets (linux-aarch64 redist):
+# build --platforms=//:my_aarch64_platform --@rules_cuda//cuda:target_gpu=on_die
 # Optional: if omitted, the highest declared redist version is used.
 build --@rules_cuda//cuda:version=13.0.0
 ```
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ use_repo(cuda, "cuda")
 
 To select CUDA versions and platforms at build time, define multiple `cuda.redist_json` entries and a single `cuda.toolkit`.
 The target CUDA redist is auto-detected from the active target platform's `@platforms//cpu` + `@platforms//os` constraints.
-Set `--platforms` to control it. linux-sbsa and linux-aarch64 share aarch64+linux constraints; the `@rules_cuda//cuda:target_gpu` flag
-disambiguates them (default `discrete` → linux-sbsa, set to `on_die` for Tegra/Jetson/Drive Thor → linux-aarch64).
+Set `--platforms` to control it. linux-sbsa and linux-aarch64 share aarch64+linux constraints; the `@rules_cuda//cuda:aarch64` flag
+disambiguates them (default `sbsa` → linux-sbsa, set to `aarch64` for Tegra/Jetson/Drive Thor → linux-aarch64).
 Use `@rules_cuda//cuda:version` to pick the CUDA version (defaults to the highest declared if unset),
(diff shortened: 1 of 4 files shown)


cloudhan:
Ideally, it indeed should be migrated to use `--platforms` as it is the principled way to do such a thing.

I am wondering is any chance we dont rely on `gpu_kind` at all?

linzhp:
The only thing `gpu_kind` is doing is disambiguating `linux-sbsa` from `linux-aarch64` (both `@platforms//cpu:aarch64 + os:linux`). For the other three platforms, `@platforms//cpu` + `@platforms//os` are sufficient on their own. Other rules ship with their own constraints to when `@platforms//cpu` + `@platforms//os` is not enough too. For example, rules_go ships with `cgo = on|off` flavors of [all OS+arch combo](https://github.com/bazel-contrib/rules_go/blob/65c1b4d0c77d67e66fca162719ac375dc96514e1/go/private/platforms.bzl#L163), with the [cgo constraint](https://github.com/bazel-contrib/rules_go/blob/65c1b4d0c77d67e66fca162719ac375dc96514e1/go/toolchain/toolchains.bzl#L61)
                                                                                                                                               
  If you'd rather not introduce a CUDA-specific constraint, I can drop `gpu_kind` and `:linux_aarch64` from auto-detection:                    
                                                                                                                                               
  - Vanilla aarch64-linux platforms would always resolve to `linux-sbsa`.                                                                      
  - Tegra/Drive users would opt in via a narrow `--@rules_cuda//cuda:target_platform=linux-aarch64` flag — kept *only* for this single case, with `--platforms` driving everything else.                                                                                                  
                                                                                
I find it a bit confusing to use `--@rules_cuda//cuda:target_platform=linux-aarch64` flag for Tegra/Drive and `--platforms` for others, but I am ok to switch to that if you prefer it over the new constraint. Let me know which direction you'd like.

cloudhan:
The flag naming is just minor issue. We can just call it `aarch64_is_aarch64` or `aarch64_is_sbsa`. 

I am thinking about a solution without messing around user defined platforms at all.
Currently, we have a flat layout of target alias
```starlark
alias(
    name = "nvcc",
    select({
        ...
        "...": ":nvcc_linux_aarch64",
        "...": ":nvcc_linux_sbsa",
        ...
    })
)
```

Consider
```starlark
alias(
    name = "nvcc",
    select({
        ...
        "...": ":nvcc_linux_sbsa_or_aarch64",  # again alias
        ...
    })
)

alias(
    name = "nvcc_linux_sbsa_or_aarch64",
    select({
        ":config_setting_is_sbsa": ":nvcc_linux_sbsa",    # true target
        ":config_setting_is_aarch": ":nvcc_linux_aarch",  # true target
    })
)
```

Then we can us a single flag without messing around platforms at all.

linzhp:
Implemented your nested-alias proposal in the latest push. `gpu_kind` constraint and the predefined `:linux_*` platforms are gone. No platform-vocabulary pollution.                                                            
                                                                                                                                                 
  **On the flag name:** I went with `--@rules_cuda//cuda:target_gpu=discrete|on_die` for richer semantics — it captures *why* the redists differ (the GPU is on PCIe vs. on-die SoC) and reads naturally for downstream users defining build configs ("I'm building for Drive Thor, which has on-die GPU"). Trade-off: it's slightly broader-scoped than your `aarch64_is_*` suggestion implies — the flag name doesn't make obvious that it's a no-op outside the aarch64-linux case.                    
                                                                                     
  If you'd rather keep the scope explicit in the flag name, happy to switch to `--@rules_cuda//cuda:aarch64_is=sbsa|aarch64` (your original) or                        
  `--@rules_cuda//cuda:aarch64_redist=sbsa|aarch64`. Both have the advantage that a user reading the flag immediately knows it only matters for aarch64. Let me know which you prefer and I'll rename — the rest of the structure stays the same.

  ---

  **Possible follow-up:** while we're cleaning up, what do you think about replacing `exec_platform` with a symmetric `exec_gpu` flag (same `discrete | on_die` values as `target_gpu`) in a separate PR?

  Pros:
  - Symmetric design — both flags do one job: aarch64 disambiguation.
  - Clean cross-compile: `--target_gpu=on_die --exec_gpu=discrete` reads more naturally than `--exec_platform=linux-sbsa`.
  - No flag with overlapping semantics vs `--platforms`.

  Cons (escape hatches lost):
  - Forcing a specific exec redist whose arch doesn't match what Bazel sees the exec platform as. Two real cases:
    - Misconfigured RBE pool (declared as wrong arch). "Fix the platform" is arguably the better answer anyway.
    - Running nvcc under emulation across architectures. Genuine edge case; could be re-added later if needed.

  Implementation note: needs a separate inner alias per component for exec (keyed on `exec_gpu`) since global flags can't be distinguished by config in `select()`.

  Happy to send that as a follow-up if you're interested, or keep `exec_platform` as-is — let me know.

charleysaintNV:
I'm confused about how the linux-sbsa or linux-aarch64 would work. The cuda redists only publish sbsa for cuda 13, is this only a cuda 12 issue?

linzhp:
As far as I know, the latest DriveOS 7.0.3/Thor only supports cuda 12.8. Do you know any on-die GPU that supports cuda 13?

charleysaintNV:
Jetpack7.x for Thor and Orin both support Cuda 13

linzhp:
OK, [here](https://developer.nvidia.com/blog/whats-new-in-cuda-toolkit-13-0-for-jetson-thor-unified-arm-ecosystem-and-more/) is the reason:

> CUDA 13.0 streamlines development for Arm platforms by unifying the CUDA toolkit across server-class and embedded devices. You no longer need to maintain separate installations or toolchains for Server Base System Architecture (SBSA) compliant servers and next-generation embedded systems like Thor.

So sasb vs aarch64 split goes away after CUDA 13. This also means not defining our own platform is the right call. People working on CUDA 13 and above will no longer need to worry about the `//cuda:target_gpu` flag: os+arch is enough to specify the target platform.

cloudhan:
So, IIUC, with linux + aarch64, in the future, for cuda components, we better just resolve to files under `linux-sbsa`?

I like the flag `--@rules_cuda//cuda:aarch64_is=sbsa|aarch64` or maybe `--@rules_cuda//cuda:aarch64=sbsa|aarch64`. Otherwise, I think this is a correct direction.

@charleysaintNV If you don't have special requirement for internal projects, I'd like to merge this PR with minor modifcation.

linzhp:
OK, updated to use `--@rules_cuda//cuda:aarch64`. 

> with linux + aarch64, in the future, for cuda components, we better just resolve to files under linux-sbsa?

Yes, linux-sbsa is the default for both exec and target platforms when OS is Linux and arch is aarch64. So future versions of CUDA will "just work"

charleysaintNV:
I think it should be fine, as long as cross compilation still works where the target platform and exec platform are still different then we should be good

PR #474, opened 2026-05-18
Inject BAZEL_CURRENT_REPOSITORY into cuda_library

AustinSchuh:
Bazel's native cc_library passes BAZEL_CURRENT_REPOSITORY as a private define so runfiles helpers (e.g. @bazel_tools//tools/cpp/runfiles) can resolve the owning repo at compile time. cuda_library skipped this, which broke any cuda TU that pulls in those headers transitively.


AustinSchuh:
https://github.com/bazelbuild/bazel/commit/1b214d2c13d4f8841c459e80ff455021b997fddb has the rules_cc piece if that is helpful.

cloudhan:
Great, previously, we need to support bazel with older version (< 7.0), many cc rule's functions cannot be used. But since the phasing out of 6.4 from bazel offical support matrix (Dec 2025), I remove the v6.4 from our CI, I think we can just safely merge it.

PR #476, opened 2026-05-22
Fix exec platform auto-detection for cross-compilation

lewissettercaci:
When cross-compiling (e.g., building for linux-aarch64 from linux-x86_64), exec tools like cicc should run on the build host platform, not the target platform. The previous constraint-based auto-detection could incorrectly select the target platform's tools in some scenarios.

This fix adds explicit host platform detection at repository instantiation time, using the detected host platform as the default for exec components. This ensures that when building with --platforms=linux-aarch64 on an x86_64 host, the x86_64 CUDA tools are selected, not the ARM64 tools.

Priority order for exec platform selection:
1. Explicit --@rules_cuda//cuda:exec_platform flag (user override)
2. Detected host platform (works for local builds & cross-compilation)
3. Constraint-based auto-detection (fallback for RBE scenarios)


cuda/platform_alias_extension.bzl
```diff
@@ -111,13 +111,60 @@ def _emit_outer_constraint_alias(build_content, alias_name, target_name, platfor
...
+    else:
+        # Unknown arch, default to x86_64
+        arch = "x86_64"
+
+    platform = "{}-{}".format(os_name, arch)
+
+    # Map to a supported CUDA platform
+    # For aarch64 on Linux, default to sbsa (server-class)
+    if platform == "linux-aarch64":
+        platform = "linux-sbsa"
```
cloudhan, line 154: I think you don't need to do the mapping for host. It should be resolved automatically by a select (`@rules_cuda//cuda:aarch64`) during analysis time.

cuda/platform_alias_extension.bzl
```diff
@@ -212,14 +269,23 @@ def _platform_alias_repo_impl(ctx):
                     build_content.append('            ":{}_{}",'.format(platform_suffix, target_name))
                 else:
                     build_content.append('            "{}",'.format(dummy_target))
-            build_content.append('        "//conditions:default": ":auto_{}",'.format(target_name))
+
+            # Default: use detected host platform
+            if host_platform in platforms_available:
+                build_content.append('        "//conditions:default": ":{}_{}",'.format(host_platform_suffix, target_name))
```
cloudhan: **[bug]** When the detected host platform is in `platforms_available` (the normal multi-platform hermetic setup), the outer exec alias hardcodes `//conditions:default` to `:{host}_{target}` and never uses `:auto_*`. Repository rules always run on the Bazel client host, not on the execution platform, so this permanently prefers host tools even when the active exec configuration is a remote/RBE worker (or any host鈮爀xec arrangement). Main's comments document constraint-based auto specifically so exec-config selects the execution platform; this change silently disables that whenever the host redist is declared.

**Suggestion:** Keep `//conditions:default": ":auto_{}"` so constraint `config_setting`s resolve against the active configuration (exec under `cfg="exec"`). If local cross-compile still picks target tools, fix the dependency edge that is missing `cfg="exec"` (or document/require `--exec_platform` for the known `:aarch64` global-flag bleed cases) rather than baking `ctx.os` into the select default.

cuda/platform_alias_extension.bzl
```diff
@@ -197,12 +244,22 @@ def _platform_alias_repo_impl(ctx):
...
 
         if platform_type == "exec":
-            # Outer alias: explicit --exec_platform flag wins; the default
-            # branch falls through to constraint-based auto-detection in
-            # the private :auto_<name> alias below.
+            # Outer alias for exec components (nvcc, cicc, etc.)
+            # Priority order:
+            #   1. Explicit --@rules_cuda//cuda:exec_platform flag
+            #   2. Detected host platform (where Bazel is running)
+            #   3. Constraint-based auto-detection (for RBE scenarios)
```
cloudhan: **[bug]** The new priority comment claims step 3 is "Constraint-based auto-detection (for RBE scenarios)," but the generated select only reaches `:auto_*` when `host_platform not in platforms_available`. Typical RBE setups declare both host and remote platforms in `cuda.redist_json`, so the host branch wins and RBE never hits constraint auto. The comment describes the intended design; the code implements the opposite for the common case.

**Suggestion:** Either restore default鈫抈:auto_*` (matching the comment and existing RBE design) or rewrite the comment and public docs to state that auto means "client host via `ctx.os`" and that RBE users must set `--@rules_cuda//cuda:exec_platform`.

cuda/platform_alias_extension.bzl
```diff
@@ -111,13 +111,60 @@ def _emit_outer_constraint_alias(build_content, alias_name, target_name, platfor
...
+        arch = "aarch64"
+    else:
+        # Unknown arch, default to x86_64
+        arch = "x86_64"
+
+    platform = "{}-{}".format(os_name, arch)
+
+    # Map to a supported CUDA platform
+    # For aarch64 on Linux, default to sbsa (server-class)
+    if platform == "linux-aarch64":
```
cloudhan: **[suggestion]** Forcing `linux-aarch64` 鈫?`linux-sbsa` at repository-rule time hardcodes the server-class redist for every physical aarch64 host. That bypasses the existing analysis-time disambiguation (`@rules_cuda//cuda:aarch64`) that main already uses for the shared aarch64+linux constraints.

**Suggestion:** Drop the host-side remap. If host detection is kept at all, emit `linux-aarch64` for aarch64 Linux and let the `:auto_*` / aarch64 inner alias + `:aarch64` flag choose sbsa vs aarch64 at analysis time.

cuda/platform_alias_extension.bzl
```diff
@@ -111,13 +111,60 @@ def _emit_outer_constraint_alias(build_content, alias_name, target_name, platfor
...
+
+    # Normalize OS name
+    if os_name.startswith("linux"):
+        os_name = "linux"
+    elif os_name.startswith("windows"):
+        os_name = "windows"
+    elif os_name.startswith("mac"):
+        os_name = "macos"
+    else:
+        # Unknown OS, default to linux
```
cloudhan: **[suggestion]** Unknown OS/arch are silently coerced to `linux` / `x86_64` instead of failing or falling through to constraint auto. A mis-detected host then gets wrong exec redists with no clear error, diverging from `cuda/private/repositories.bzl` `_detect_platform`, which fails on unsupported OS/arch.

**Suggestion:** Prefer falling through to `:auto_*` (or `fail()` with the raw `ctx.os.name` / `ctx.os.arch`) on unknown values.

cuda/platform_alias_extension.bzl
```diff
@@ -197,12 +244,22 @@ def _platform_alias_repo_impl(ctx):
         )
 
         if platform_type == "exec":
-            # Outer alias: explicit --exec_platform flag wins; the default
-            # branch falls through to constraint-based auto-detection in
-            # the private :auto_<name> alias below.
+            # Outer alias for exec components (nvcc, cicc, etc.)
```
cloudhan: **[suggestion]** The multi-line priority-order block restates what the following select already encodes and embeds design rationale better kept in short docs next to `exec_platform`. It also drifts from the code (claims RBE uses constraint auto while host-available path never reaches it).

**Suggestion:** Collapse to a brief comment on the non-obvious constraint and keep behavior documented in `//cuda:BUILD.bazel` / README.

cuda/platform_alias_extension.bzl
```diff
@@ -111,13 +111,60 @@ def _emit_outer_constraint_alias(build_content, alias_name, target_name, platfor
     build_content.append(")")
     build_content.append("")
 
+def _detect_host_platform(ctx):
```
cloudhan: **[nit]** `_detect_host_platform`'s docstring is longer than the function and mostly restates the body.

**Suggestion:** One-liner, e.g. "Map `ctx.os` to a `SUPPORTED_PLATFORMS` key for the exec-alias default."

PR #488, opened 2026-08-06
feat: move CUDA RDC device runtime libs into the toolchain

charleysaintNV:
# Move CUDA RDC device runtime libs into the toolchain

## Summary

This change removes unconditional propagation of CUDA device runtime static archives from the normal CUDA runtime dependency and moves them into the CUDA toolchain.

The goal is to let non-RDC consumers of `@rules_cuda//cuda:runtime` keep linking against shared `libcudart` without also pulling `libcudadevrt.a` / `libculibos.a` into downstream shared libraries. RDC builds still receive those static archives from the toolchain for device-link and final host-link correctness.

## Changes

- Adds `device_runtime_static_libs` to `CudaToolkitInfo`.
- Discovers local toolkit `libcudadevrt.a` / `libculibos.a` and generated redist aliases as CUDA toolchain inputs.
- Adds the device runtime static archives to CUDA device-link actions.
- Propagates device runtime static archives into final C++ linking only for `cuda_library(rdc = True)`.
- Removes `cudadevrt_a` / `culibos_a` from the normal dynamic `cuda_runtime` dependency path on Linux.
- Preserves Windows dynamic-runtime link behavior, where CUDA’s runtime import-library interface includes `cudadevrt.lib`.
- Exposes device-runtime component targets needed by generated toolchains.
- Handles CUDA 13+ `culibos` ownership for both local toolkits and redist-generated alias repos so `culibos_a` is emitted once and resolves to the component that owns the archive.

## Validation

Rules CUDA:

```sh
bazel test //tests/flag:cuda_library_dlink_copts_and_host_copts_flag_test
bazel build @rules_cuda_examples//rdc:main_from_library
bazel build @rules_cuda_examples//basic:main
```

Link-shape checks:

```sh
bazel aquery --noshow_progress 'mnemonic("CppLink", @rules_cuda_examples//basic:main)' --output=text
bazel aquery --noshow_progress 'mnemonic("CppLink", @rules_cuda_examples//rdc:main_from_library)' --output=text
```

- Non-RDC `basic:main` final link uses shared `libcudart` and does not include `libcudadevrt.a`, `libculibos.a`, or `libcudart_static.a`.
- RDC `rdc:main_from_library` final link includes `libcudadevrt.a` and `libculibos.a` from the CUDA toolchain.

Downstream smoke validation:

```sh
bazel build --override_module=rules_cuda=/home/csaint/nv/csaint/rules_cuda \
  :visual_global_localization_node_so
```

[...]

PR #490, opened 2026-08-11
fix: report the selected CUDA version, not the maximum declared one

### LINKED ISSUE #477: Multi-versioning fails for CUDA toolkit
For the autogenerated NVCC toolkit BUILD stubs, there is inconsistency between the declared and actual version of the toolkit components. This is due to the highest version being declared for the toolkit info, while all the other variables are aliases that get selected as a function of version.

I do see that this was done intentionally, as mentioned in this snippet here:
`cuda/extensions.bzl`:
```bazel
    for _, toolkit in registrations.items():
        if components_mapping != None:
            # Always use the maximum version so the toolkit includes all components.
            # Components that don't exist in older versions will fall back to dummy.
            toolkit_version = sorted(redist_versions, key = _version_sort_key)[-1]
            cuda_toolkit(name = toolkit.name, components_mapping = components_mapping, version = toolkit_version)
        else:
            cuda_toolkit(**_module_tag_to_dict(toolkit))
```

However, issues arise when using a 12.X and 13.X version due to false positives being set for certain compiler flags. In particular, I was seeing a failure when using CUDA 12.X with version due to this random seed feature flag in particular (which, as stated in the comment, fails on CUDA <= 12.9):

`cuda/private/toolchain_configs/nvcc.bzl`:
```bazel
    # NOTE: this only works on compiler newer than 12.9
    nvcc_fixed_random_seed_feature = feature(
        name = "nvcc_fixed_random_seed",
        enabled = True,
        flag_sets = [
            flag_set(
                actions = [
                    ACTION_NAMES.cuda_compile,
                    ACTION_NAMES.device_link,
                ],
                flag_groups = [flag_group(flags = ["--frandom-seed=%{output_file}"])],
            ),
        ] if nvcc_version_ge(ctx, 12, 9) else [],
    )
```

A minimal setup to see the issue is to declare a 12.X and 13.X distribution, and attempt to build with the 12.X version:

`MODULE.bazel`:
```bazel
bazel_dep(name = "rules_cuda")

git_override(
    module_name = "rules_cuda",
    commit = "ccd93bdfb002172e570993846db16fb6b8670136",  # < This is the commit I've been working off as I said - the issue should still exist on main from what I've seen
    remote = "https://github.com/bazel-contrib/rules_cuda.git",
)

cuda.redist_json(
    name = "cuda_12_6_0",
    platforms = [
        "linux-aarch64",
        "linux-sbsa",
        "linux-x86_64",
    ],
    version = "12.6.0",
)
cuda.redist_json(
    name = "cuda_13_0_1",
    platforms = [
        "linux-sbsa",
        "linux-x86_64",
    ],
    version = "13.0.1",
)
cuda.toolkit(
    name = "cuda",
)
use_repo(cuda, "cuda")
```

`BUILD.bazel`:

```
load("@bazel_skylib//rules:write_file.bzl", "write_file")
load("@rules_cuda//cuda:defs.bzl", "cuda_library")

write_file(
    name = "dummy",
    out = "dummy.cu",
    content = ["#include <cuda_runtime.h>"],
)

cuda_library(
    name = "dummy_lib",
    srcs = [":dummy"],
    deps = ["@cuda//:cuda_runtime"],
)
```

Build command:

```bash
bazel build --@rules_cuda//cuda:enable=true --@rules_cuda//cuda:version=12.6.0 //:dummy_lib
```

Error:

```
INFO: Analyzed target //:dummy_lib (84 packages loaded, 5855 targets configured).
ERROR: /<>/BUILD.bazel:55:13: Compiling bazel-out/aarch64-fastbuild/bin/dummy.cu failed: (Exit 1): nvcc failed: error executing CudaCompile command (from cuda_library rule target //:dummy_lib) external/rules_cuda++toolchain+cuda_nvcc_linux_sbsa_12_6_0/nvcc/bin/nvcc -x cu -Xcompiler -fPIC -ccbin /usr/bin/clang -I . -I bazel-out/aarch64-fastbuild/bin -I ... (remaining 37 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
nvcc fatal   : Unknown option '--frandom-seed=bazel-out/aarch64-fastbuild/bin/_objs/dummy_lib/dummy.pic.o'
Target //:dummy_lib failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 5.339s, Critical Path: 0.08s
INFO: 2 processes: 3 action cache hit, 2 internal.
ERROR: Build did NOT complete successfully
```

The natural fix would be to add the version as a `select` statement as done for the other versions. As constructed, this does technically need to be declared in the toolchain `BUILD` file itself since the version attribute(s) accepts a string or integer depending on the rule. The attached patch addresses the issue (based off commit `ccd93bdfb002172e570993846db16fb6b8670136`): [0000_toolchain_build_version_select.patch](https://github.com/user-attachments/files/28274842/0000_toolchain_build_version_select.patch)

> [!NOTE]
> This patch was AI generated so there's a bit of slop / unnecessary stuff, but I think it mostly illustrates the high level changes that are required.

janickm:
Fixes #477.

## Problem

With several `cuda.redist_json` declared, the component aliases resolve through `--@rules_cuda//cuda:version`, but the generated toolchain hardcodes the **maximum** declared version:

```starlark
# Always use the maximum version so the toolkit includes all components.
toolkit_version = sorted(redist_versions, key = _version_sort_key)[-1]
```

Version-gated behaviour is then decided for a version other than the nvcc that actually runs. Declaring 13.0.2 and 12.8.1 and building at 12.8.1 gets `--frandom-seed`, which is guarded by `nvcc_version_ge(ctx, 12, 9)` and read 13.0 from the stamped attributes:

```
nvcc fatal   : Unknown option '--frandom-seed=bazel-out/k8-fastbuild/bin/_objs/kernel_lib/kernel.pic.o'
```

## Fix

`version`, `nvcc_version_major` and `nvcc_version_minor` now `select()` on `cuda:version`, mirroring what `platform_alias_repo` already does for the component labels.

The selects are generated in the BUILD templates rather than passed through Starlark, because `cuda_toolkit_info.version` is a string attribute while `nvcc_version_major`/`_minor` are ints, so there is no single value to hand down.

**`cuda_toolkit_info.version` is included deliberately, not just the nvcc attributes.** It becomes `CudaToolkitInfo.version_major`, which picks `--image3` over `--image` for CUDA >= 13 in `dlink.bzl`. Fixing only the nvcc attributes would leave a 12.x build under a 13.x maximum device-linking with CUDA 13 syntax.

`//conditions:default` keeps the maximum declared version, matching what the component aliases already resolve to when `cuda:version` is unset.

## Single-version output is unchanged

No `config_setting` and no `select` are emitted for a single declared version. Diffing the generated `@cuda//toolchain/BUILD` for `tests/integration/toolchain_redist_json` before and after, the only change is the literals no longer needing the string-substitution workaround:

```diff
-    # int("%{foo}") instead of %{foo} to make the file valid syntactically.
-    nvcc_version_major = int("12"),
-    nvcc_version_minor = int("6"),
+    nvcc_version_major = 12,
+    nvcc_version_minor = 6,
```

## Test

`tests/integration/toolchain_redist_json_multi` declares 11.7.0 and 12.6.3, both below the 12.9 gate, so no version-gated flag differs between them and this cannot surface there. The new `toolchain_redist_json_version_gate` case declares 13.0.2 and 12.8.1, which straddle that gate, and compiles a `cuda_library` against each.

Two things worth flagging about it:

- **It must not run with `CUDA_REDIST_VERSION_OVERRIDE` set.** CI exports it for the whole script, and it rewrites the version of *every* `redist_json`, collapsing both declarations onto one version and leaving nothing multi-version to test. The case unsets it for its own builds, so it stays meaningful under the existing CI invocation. This is also why the existing multi-version case degenerates to single-version in CI today.
- `crt` and `nvvm` are requested only from 13.0.2, since they are required from CUDA 13 on and are not published before it.

## Verification

On this branch, `--@rules_cuda//cuda:version` now drives both the nvcc that runs and the flags it is given:

| `cuda:version` | nvcc invoked | `--frandom-seed` |
| --- | --- | --- |
| 13.0.2 | `cuda_nvcc_linux_x86_64_13_0_2` | present |
| 12.8.1 | `cuda_nvcc_linux_x86_64_12_8_1` | absent |

Both `cuda_library` builds succeed; on `main` the 12.8.1 one fails with the error above.

Also run locally:

- `bazel test //tests/...` — 58 pass, 6 skipped
- `examples`: `//basic:all //rdc:all //if_cuda:main` — including the `rdc` device-link path that exercises the `version_major` gate
- `tests/integration/test_all.sh` for the cases this machine can run (root, none, collision, redist_json, and the new case) — exit 0
- `pre-commit run --all-files` — buildifier and prettier pass; commitizen passes on both commit messages

Two integration builds could not run here for environmental reasons unrelated to this change: CUDA 11.7 does not accept this host's gcc 13 / glibc headers (`identifier "_Float32" is undefined`), which reproduces identically on unpatched `main`. The 12.6.3 half of that case passes.

I have also run this against a real repository that declares CUDA 12.8.1 and 13.2.2 side by side and builds torch CUDA extensions against both. It previously needed `cuda_features = ["-nvcc_fixed_random_seed"]` on every 12.8 target; with this change that workaround is removable, each dependency set resolves its own nvcc, and the flag is applied only where the nvcc supports it.

## Note on the patch in #477

The reporter attached a patch and noted it was AI-generated with some slop in it. I wrote this independently and did not carry that patch over; the two agree on the overall approach, which seemed like the natural fix. Credit for the diagnosis is theirs.

### DIFF
--- a/cuda/private/repositories.bzl
+++ b/cuda/private/repositories.bzl
@@ -349,6 +349,12 @@ cuda_toolkit = repository_rule(
         "nvcc_version": attr.string(
             doc = "nvcc version. Required for deliverable toolkit only. Fallback to version if omitted.",
         ),
+        "toolkit_versions": attr.string_list(
+            doc = "All cuda toolkit versions reachable through the component aliases, i.e. every " +
+                  "version declared with `cuda.redist_json`. When more than one is present, the " +
+                  "generated toolchain selects among them on @rules_cuda//cuda:version instead of " +
+                  "hardcoding `version`.",
+        ),
     },
     configure = True,
     local = True,
--- a/cuda/extensions.bzl
+++ b/cuda/extensions.bzl
@@ -248,10 +248,22 @@ def _impl(module_ctx):
 
     for _, toolkit in registrations.items():
         if components_mapping != None:
+            sorted_redist_versions = sorted(redist_versions, key = _version_sort_key)
+
             # Always use the maximum version so the toolkit includes all components.
             # Components that don't exist in older versions will fall back to dummy.
-            toolkit_version = sorted(redist_versions, key = _version_sort_key)[-1]
-            cuda_toolkit(name = toolkit.name, components_mapping = components_mapping, version = toolkit_version)
+            toolkit_version = sorted_redist_versions[-1]
+
+            # All declared versions are passed along so that the generated toolchain can
+            # select the version matching @rules_cuda//cuda:version. Without it the
+            # toolchain would report `toolkit_version` regardless of which nvcc the
+            # component aliases actually resolve to.
+            cuda_toolkit(
+                name = toolkit.name,
+                components_mapping = components_mapping,
+                version = toolkit_version,
+                toolkit_versions = sorted_redist_versions,
+            )
         else:
             cuda_toolkit(**_module_tag_to_dict(toolkit))
 
--- a/cuda/private/template_helper.bzl
+++ b/cuda/private/template_helper.bzl
@@ -18,6 +18,108 @@ def _expand_template(repository_ctx, tpl_label, substitutions):
         template_content = template_content.replace(k, v)
     return template_content
 
+def _version_label(version):
+    return version.replace(".", "_").replace("-", "_")
+
+def _major_minor(version):
+    """Split a version into its (major, minor) pair, or None if it has neither.
+
+    Args:
+        version: A version string such as "12.8.1" or "11.5.1-1ubuntu1".
+
+    Returns:
+        A (major, minor) tuple of strings, or None if `version` is not of that shape.
+    """
+    parts = version.split("-", 1)[0].split(".")
+    if len(parts) < 2 or not parts[0].isdigit() or not parts[1].isdigit():
+        return None
+    return (parts[0], parts[1])
+
+def _versions_to_select_on(toolkit_versions):
+    """The versions a toolchain attribute can be selected on.
+
+    A single version needs no select at all, and one whose major/minor cannot be parsed
+    has nothing to compare against `--@rules_cuda//cuda:version`, so both are dropped.
+
+    Args:
+        toolkit_versions: Every version declared with `cuda.redist_json`.
+
+    Returns:
+        The versions to emit a select branch for, or [] to keep the attribute a literal.
+    """
+    if not toolkit_versions or len(toolkit_versions) < 2:
+        return []
+    return [v for v in toolkit_versions if _major_minor(v) != None]
+
+def _version_config_settings(toolkit_versions):
+    """config_settings matching `--@rules_cuda//cuda:version`, one per declared version.
+
+    Emitted into the same package as the selects that read them.
+
+    Args:
+        toolkit_versions: The versions returned by `_versions_to_select_on`.
+
+    Returns:
+        The config_setting definitions, or "" when no select is needed.
+    """
+    if not toolkit_versions:
+        return ""
+
+    lines = []
+    for version in toolkit_versions:
+        lines.extend([
+            "config_setting(",
+            '    name = "toolkit_version_is_{}",'.format(_version_label(version)),
+            '    flag_values = {{"@rules_cuda//cuda:version": "{}"}},'.format(version),
+            ")",
+            "",
+        ])
+    return "\n".join(lines).rstrip("\n")
+
+def _substitute_version_config_settings(substitutions, toolkit_versions):
+    """Add the config_setting substitution, keyed so an empty value leaves no blank line.
+
+    The placeholder sits on its own line in the template. Substituting an empty string
+    would leave that line blank, so when there is nothing to emit the surrounding newlines
+    are consumed along with it.
+
+    Args:
+        substitutions: The substitution dict to add to.
+        toolkit_versions: The versions returned by `_versions_to_select_on`.
+    """
+    config_settings = _version_config_settings(toolkit_versions)
+    if config_settings:
+        substitutions["# %{toolkit_version_config_settings}"] = config_settings
+    else:
+        substitutions["\n# %{toolkit_version_config_settings}\n"] = ""
+
+def _version_select(attr_name, toolkit_versions, value_of_version, default_value):
+    """An attribute assignment selecting its value on the active CUDA version.
+
+    Args:
+        attr_name: Name of the attribute to assign.
+        toolkit_versions: The versions returned by `_versions_to_select_on`.
+        value_of_version: Callback mapping a version to its already-quoted attribute value.
+        default_value: Already-quoted value used when `cuda:version` is unset, which
+            leaves the toolchain reporting the same version the component aliases
+            resolve to by default (the maximum declared one).
+
+    Returns:
+        Either `attr = <default>,` or `attr = select({...}),`.
+    """
+    if not toolkit_versions:
+        return "{} = {},".format(attr_name, default_value)
+
+    lines = ["{} = select({{".format(attr_name)]
+    for version in toolkit_versions:
+        lines.append('    ":toolkit_version_is_{}": {},'.format(
+            _version_label(version),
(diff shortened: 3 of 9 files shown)


cuda/private/repositories.bzl
```diff
@@ -228,7 +228,7 @@ def config_cuda_toolkit_and_nvcc(repository_ctx, cuda):
     template_helper.generate_defs_bzl(repository_ctx, cuda.version_major, cuda.version_minor, is_local_ctk == True)
 
     # Generate @cuda//toolchain/BUILD
-    template_helper.generate_toolchain_build(repository_ctx, cuda)
+    template_helper.generate_toolchain_build(repository_ctx, cuda, repository_ctx.attr.toolkit_versions)
```
cloudhan: This new arg make no sense when you have `repository_ctx` already.

janickm: good point, sorry for this blunder :facepalm: updated

janickm:
[seems like the CI needs to be triggered with an approval, haven't been able to verified this remotely yet]

cuda/private/repositories.bzl
janickm: [resolving as addressed / minor]

janickm:
Force-pushed a fix for the CI failures; they were all in the new integration case, not the change itself.

Three things were wrong with it:
- It requested a subset of `components`, but `cudart` depends on other components' header targets (`curand_headers`), which broke the clang jobs. It now takes all components, like the existing multi-version case.
- It declared only `linux-x86_64`, so on Windows - where `test_all.sh` pins `exec_platform=windows-x86_64` - every tool resolved to the dummy. It now declares `windows-x86_64` too.
- It paired 13.0.2 with 12.8.1. Crossing 12 → 13 moves `libdevice` into the `nvvm` component, which the clang toolchain resolves from the maximum declared version, and clang also targets a PTX ISA the older ptxas rejects. The pair is now 12.9.1 / 12.8.1, which straddles the `--frandom-seed` gate within one major, and the case pins `cuda:compiler=nvcc` since the flag under test is an nvcc flag. I confirmed it still reproduces the original failure without the fix.

I ran the full matrix on a branch in my fork: everything passes except the two windows-2025 jobs, which fail identically on `main` at `@rules_cuda_examples//basic:all` with the MSVC `aligned_storage` assert that #487 addresses.

Separately, the `libdevice` behaviour above looks like the same root cause in the clang path: `nvvm_all_files` and `--cuda-path` are chosen from the maximum declared version at repository-evaluation time, so it can't use a `select()`. I left it out of this PR - happy to file it separately if you'd like it tracked.

cloudhan:
Please rebase onto latest main.

janickm:
Rebased onto b283e98.

#488 touched the same three toolchain templates, so there were two conflicts, both the same shape: it moved ptxas up next to path while this replaces `version = "%{cuda_version}"` with a placeholder. Kept both. `device_runtime_static_libs` and the `version` selects coexist fine in the generated toolchain:
```
device_runtime_static_libs = ["@cuda_cudart//:cudadevrt_a", "@cuda_cudart//:culibos_a"],
version = select({...}),
nvcc_version_major = select({...}),
```
Re-checked after the rebase: `//tests/...` (82 pass), examples including `//rdc:all` since #488 reworked device linking and this changes the `version_major` that gates `--image3`, `test_all.sh`, and pre-commit. Single-version toolchain output still differs from main only in `nvcc_version_major/_minor` becoming plain ints. I also re-confirmed against a clean b283e98 that the new case still fails without the fix, so the rebase didn't turn it into a no-op.

The full matrix is green now on a branch in my fork, all 10 jobs including both windows-2025 ones - #491 fixed those, so the failures I mentioned earlier are gone and nothing is outstanding.

Issue #372, opened 2025-06-27
Sandbox absolute paths leaking into binary


kirschem-fernride:
Hey,

we're using `rules_cuda` with these flags:

```
build --@rules_cuda//cuda:compiler=nvcc
build --@rules_cuda//cuda:archs="compute_86:compute_86,sm_86;compute_75:compute_75,sm_75"
build --@rules_cuda//cuda:copts="--std=c++17"
build --@rules_cuda//cuda:copts=-allow-unsupported-compiler
build --@rules_cuda//cuda:copts=-Xcompiler=-Wall
build --@rules_cuda//cuda:copts=-Xcompiler=-Wextra
build --@rules_cuda//cuda:copts=-Xcompiler=-Werror
build --@rules_cuda//cuda:copts=-Werror=all-warnings
build --@rules_cuda//cuda:runtime=@cudart//:cuda_runtime
build --@rules_cuda//cuda:host_copts=-ftrapping-math
build --@rules_cuda//cuda:copts=-Xcompiler=-ffile-prefix-map=$(pwd)=.
build --@rules_cuda//cuda:copts=-Xcompiler=-fdebug-prefix-map=$(pwd)=.
```

The compiled object files of `cuda_library` targets contain absolute paths in the `.nv_fatbin` section to the bazel sandbox used for the build.
This is a problem for reproducibility of the build.

This problem happens ONLY on cuda targets. 
This helped for some of the paths:
```
build --@rules_cuda//cuda:copts=-Xcompiler=-ffile-prefix-map=$(pwd)=.
build --@rules_cuda//cuda:copts=-Xcompiler=-fdebug-prefix-map=$(pwd)=.
```
but not all of them.

One example diff excerpt between two builds:

```
    -/tmp/tmpg05fa8va_repro_check_1/sandbox/linux-sandbox/4/execroot/_main/src/edacted/abc_utils
    -/tmp/tmpg05fa8va_repro_check_1/sandbox/linux-sandbox/4/execroot/_main/external/_main~_repo_rules~cudart/include
    +/tmp/tmpyvmnqnpq_repro_check_2/sandbox/linux-sandbox/4/execroot/_main/src/redacted/abc_utils
    +/tmp/tmpyvmnqnpq_repro_check_2/sandbox/linux-sandbox/4/execroot/_main/external/_main~_repo_rules~cudart/include
```

Any ideas where this does come from?

cloudhan:
[...]

PR #494, opened 2026-08-18
Place obj file in seperate dir

comicfans:
without --objdir-as-tempdir,  nvcc will put temp cpp file under shared /tmp directory, while multi compile running in parallel, they create non-deterministic file. using 
```bash
llvm-nm -a cuda_library.a |grep tmp
```
it might shows 
`tmpxft_00000002_00000000-6_kernel.cudafe1.cpp` or 
`tmpxft_00000002_00000001-6_kernel.cudafe1.cpp`

apply [--objdir-as-tempdir](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#objdir-as-tempdir-objtemp) arguments fix this by place stable file under object directory ,  it explicitly declared same file compiled with different options shouldn't be in same directory. rules_cuda now placed object file and pic object file under same directory, which might break --objdir-as-tempdir.

this PR address this by place output under different category folder.

Note: test is written by codex, revert place-file-under-category commit confirms the test failed as expected.

### DIFF
--- a/cuda/private/actions/compile.bzl
+++ b/cuda/private/actions/compile.bzl
@@ -63,13 +63,21 @@ def compile(
         filename = None
         filename = cuda_helper.get_artifact_name(cuda_toolchain, artifact_category_name, basename)
 
-        # Objects are placed in <_prefix>/<tgt_name>/<filename>.
+        # Keep action output variants in separate directories because NVCC may place 
+        # temporary files beside the output object, see 
+        # https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#objdir-as-tempdir-objtemp
+        # ... same file is compiled with different options ...
+        # Compiling them in the same directory can either cause the compilation to fail or produce incorrect results.
+        #
+        # Objects are placed in <_prefix>/<tgt_name>/<artifact_category>/<filename>.
         # For files with the same basename, say srcs = ["kernel.cu", "foo/kernel.cu", "bar/kernel.cu"], we get
-        # <_prefix>/<tgt_name>/0/kernel.<ext>, <_prefix>/<tgt_name>/1/kernel.<ext>, <_prefix>/<tgt_name>/2/kernel.<ext>.
+        # <_prefix>/<tgt_name>/object_file/0/kernel.<ext>,
+        # <_prefix>/<tgt_name>/pic_object_file/0/kernel.<ext>,
+        # <_prefix>/<tgt_name>/pic_object_file/1/kernel.<ext>.
         # Otherwise, the index is not presented.
         if basename_counter[basename] > 1:
             filename = "{}/{}".format(basename_index, filename)
-        obj_file = actions.declare_file("{}/{}/{}".format(_prefix, ctx.attr.name, filename))
+        obj_file = actions.declare_file("{}/{}/{}/{}".format(_prefix, ctx.attr.name, artifact_category_name, filename))
         ret.append(obj_file)
 
         var = cuda_helper.create_compile_variables(
--- a/tests/flag/BUILD.bazel
+++ b/tests/flag/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@rules_cuda//cuda:defs.bzl", "cuda_library")
+load(":compile_output_test.bzl", "cuda_compile_output_directory_test")
 load(
     ":flag_validation_test.bzl",
     "cuda_library_c_dbg_flag_test",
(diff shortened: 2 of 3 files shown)


cloudhan:
See https://github.com/bazel-contrib/rules_cuda/issues/372#issuecomment-3445430882

IIRC, objtemp is not mandatory to achieve reproducible build. The key is the random seed thing.

comicfans:
> See [#372 (comment)](https://github.com/bazel-contrib/rules_cuda/issues/372#issuecomment-3445430882)
> 
> IIRC, objtemp is not mandatory to achieve reproducible build. The key is the random seed thing.

Thanks for the info.  seems nvcc_fixed_random_seed_feature only works with cuda > 12.9 ? I'm working with 12.6 so not sure how this affect build behavior. according to https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#frandom-seed-frandom-seed it affect symbol/variable name generation. and --objdump affect temp file path embedded in achive file so both should be required. for my 12.6 cuda environment,  set  `--objtemp` and `-Xcompiler=-fdebug-compilation-dir=.` flags gives same hash. but this might not as comprehensive as #372

cloudhan:
prior to 12.9, there is no way to stably make the build reproducible, IIUC. Because nvcc almost always encode full path info in the final artifact, and objtemp itself is not stable...

comicfans:
> prior to 12.9, there is no way to stably make the build reproducible, IIUC. Because nvcc almost always encode full path info in the final artifact, and objtemp itself is not stable...

 I did a quick check by running strings ( grep '/' and cpp) on the generated library.a and looking for embedded absolute paths. before adding `--objtemp` and `-Xcompiler=-fdebug-compilation-dir=.` my cuda library.a contains following strings:

```
/               0           0     0     0       538       `
//                                              20        `
fused_kernel.pic.o/
/0              0           0     0     644     63824     `
clang version 20.1.0 (https://github.com/llvm/llvm-project 24a30daaa559829ad079f2ff7f73eb4e18095f88)
/tmp/tmpxft_00000002_00000000-6_fused_kernel.cudafe1.cpp
/home/xinyuwang/.cache/bazel/_bazel_xinyuwang/06ac47915257f6c9e982f482518f8fc1/sandbox/linux-sandbox/696/execroot/_main
clang version 20.1.0 (https://github.com/llvm/llvm-project 24a30daaa559829ad079f2ff7f73eb4e18095f88)
/home/xinyuwang/.cache/bazel/_bazel_xinyuwang/06ac47915257f6c9e982f482518f8fc1/sandbox/linux-sandbox/696/execroot/_main
proj/dir/csrc
external/rules_cuda++toolchain+cuda_cudart_v12.6.77/cudart/include
/tmp
external/rules_cuda++toolchain+cuda_nvcc_v12.6.85/nvcc/include/crt
/tmp/tmpxft_00000002_00000000-6_fused_kernel.cudafe1.cpp
tmpxft_00000002_00000000-6_fused_kernel.cudafe1.cpp


```

After adding those options, the corresponding paths became relative:

```
/               0           0     0     0       538       `
//                                              20        `
fused_kernel.pic.o/
/0              0           0     0     644     63808     `
clang version 20.1.0 (https://github.com/llvm/llvm-project 24a30daaa559829ad079f2ff7f73eb4e18095f88)
bazel-out/k8-fastbuild/bin/proj/dir/_objs/fused_kernel/pic_object_file/fused_kernel.pic.o.cudafe1.cpp
clang version 20.1.0 (https://github.com/llvm/llvm-project 24a30daaa559829ad079f2ff7f73eb4e18095f88)
proj/dir/csrc
external/rules_cuda++toolchain+cuda_cudart_v12.6.77/cudart/include
bazel-out/k8-fastbuild/bin/proj/dir/_objs/fused_kernel/pic_object_file
external/rules_cuda++toolchain+cuda_nvcc_v12.6.85/nvcc/include/crt
bazel-out/k8-fastbuild/bin/proj/dir/_objs/fused_kernel/pic_object_file/fused_kernel.pic.o.cudafe1.cpp
fused_kernel.pic.o.cudafe1.cpp


```

IIUC, the bazel-out/k8-fastbuild/bin prefix should remain stable as long as the same configuration is used, so these paths should be treated as deterministic, or did I miss something ?

When you say that objtemp itself is not stable prior to CUDA 12.9, do you mean that using -objtemp may cause nvcc to crash or produce incorrect output, or that it can still generate nondeterministic artifacts under some conditions?

cloudhan:
it can still generate nondeterministic artifacts.

PR #496, opened 2026-08-24
fix: resolve CCCL headers from the CUDA 13 include path

### LINKED ISSUE #489: cccl_headers uses the pre-CUDA-13 include path, so <thrust/...> does not resolve on CUDA 13
On a CUDA 13 toolkit, `@cuda//:cccl_headers` sets `includes = ["cccl/include"]`, but CUDA 13 moved the CCCL headers one level deeper, to `cccl/include/cccl/`. Anything including `<thrust/complex.h>` (or `<cub/...>`, `<cuda/...>`, `<nv/...>`) through that target fails to find the header.

The per-library targets in `cuda/private/templates/BUILD.cccl` already handle this correctly:

```starlark
includes = if_cuda_toolkit_version_ge(
    (13, 0),
    ["%{component_name}/include/cccl"],
    ["%{component_name}/include"],
),
```

But `cccl_headers` is not generated from that fragment. It comes from the generic `cuda/private/templates/BUILD.dctk_comp`, which is version-unaware and always emits the shallow path:

```starlark
cc_library(
    name = "%{component_name}_headers",
    hdrs = [":%{component_name}_header_files"],
    includes = ["%{component_name}/include"],
)
```

This also reaches `@cuda//:cuda_headers`, since `BUILD.dctk_cuda` aggregates every `<component>_headers` target.

The `hdrs` glob (`cccl/include/**`) does pick the files up, so the headers are present -- only the include path is wrong. `#include <cccl/thrust/complex.h>` resolves; `#include <thrust/complex.h>` does not.

Observed with a `cuda.redist_json` toolkit at 13.2.2 on `linux-x86_64`, in the generated `@cuda_cccl_linux_x86_64_13_2_2//:cccl_headers`. Depending on `@cuda//:thrust` instead is a working workaround, since that target is version-aware.

This looks independent of the multi-version work -- the generic fragment is version-unaware for a single declared version too -- but I have only tested the redist path, not a local CTK install.

Happy to send a PR if you'd like it fixed in `BUILD.dctk_comp` (guarding the generic target) versus having the `cccl` fragment override its own `_headers` target -- I don't want to guess which layering you'd prefer.

janickm:
Fixes #489.

## Problem

`cccl_headers` is generated from the generic `<component>_headers` in `BUILD.dctk_comp`, which always sets `includes` to `<component>/include`. CUDA 13 moved the CCCL headers one level deeper, into `include/cccl/`, so that path resolves `<cccl/thrust/...>` but not `<thrust/...>`:

```
fatal error: thrust/complex.h: No such file or directory
```

The per-library targets in `BUILD.cccl` already handle the move, but `cccl_headers` is not generated from that fragment. It also reaches `cuda_headers`, which aggregates every component's headers target, so both fail the same way.

## Fix

Per your suggestion in the issue: `BUILD.cccl` now declares `cccl_headers` itself, using the same `if_cuda_toolkit_version_ge` as the neighbouring `thrust`/`cub`/`libcudacxx` targets, and the generic target in `BUILD.dctk_comp` is suppressed for that component. The two fragments are concatenated into one BUILD file, so only one of them may declare it — the suppression is required, not cosmetic. It follows the shape of `_component_owns_cuda_repo_alias` from #488.

Keeping the version knowledge in the fragment that already has it avoids teaching the generic template about CCCL's layout.

One wrinkle worth flagging: the new target is confined to deliverable repos via `if_local_cuda_toolkit`. A local toolkit expands `BUILD.cccl` into `@cuda` itself, where `%{component_name}` is `cuda` and `BUILD.lctk_cuda` already declares `cuda_headers` — without the guard, `toolchain_root` fails with a duplicate `cuda_headers`.

## Verification

Non-cccl component repos generate byte-identical BUILD files, and exactly one `cccl_headers` is declared:

| Toolkit | `cccl_headers` resolves to |
| --- | --- |
| CUDA 13.0.2 | `cccl/include/cccl` |
| CUDA 12.8.1 | `cccl/include` |

The new integration case builds a kernel including `<thrust/complex.h>` through both `cccl_headers` and the `cuda_headers` aggregate. It declares a single CUDA version, so `CUDA_REDIST_VERSION_OVERRIDE` sweeps it across the versions in the matrix and covers both layouts. I confirmed it fails on unpatched main and passes here.

The targets pass nvcc's own `-std=c++17`, not an `-Xcompiler` host flag: thrust refuses to compile below C++17, some CI hosts default lower, and a host-only flag leaves nvcc compiling the CCCL headers as C++14 (which fails on MSVC with "inline specifier allowed on function declarations only").

Full matrix green on a branch in my fork — all 10 `Test Example Build` jobs including the three Windows ones, plus `Test Utilities` and `pre-commit`. Also ran `bazel test //tests/...` (82 pass), the examples including `//rdc:all`, and `test_all.sh`.

## Separate, not fixed here

While testing this across two declared versions I hit a different instance of the max-version problem: every component repo loads `@cuda//:defs.bzl`, whose `ctk_version` is stamped from the maximum declared version, so `if_cuda_toolkit_version_ge` returns the wrong answer inside a non-maximum component repo. On current main, `@cuda_cccl_..._12_8_1//:thrust` already resolves `cccl/include/cccl` when 13.x is also declared.

That is pre-existing and affects `thrust`, `cub` and `libcudacxx` equally, and it cannot be fixed in the fragments, since `if_cuda_toolkit_version_ge` is a load-time function reading a stamped constant. It needs each component repo to know its own version, which the `version` attribute already carries. I kept it out of this PR and will file it separately unless you would rather it were folded in.

### DIFF
--- a/cuda/private/template_helper.bzl
+++ b/cuda/private/template_helper.bzl
@@ -149,10 +149,19 @@ def _expand_dctk_cuda(repository_ctx, components):
     }
     return _expand_template(repository_ctx, tpl_label, substitutions = substitutions)
 
+# Components whose own fragment declares `<component>_headers`, so BUILD.dctk_comp must
+# not: both fragments land in one BUILD file, and only one may declare a given target.
+# cccl does, because CUDA 13 moved its headers into an `include/cccl/` subdirectory and
+# only its fragment knows to follow that.
+_COMPONENTS_DECLARING_OWN_HEADERS = ["cccl"]
+
 def _expand_dctk_component(repository_ctx, component):
     tpl_label = Label("//cuda/private:templates/BUILD.dctk_comp")
     substitutions = {
         "%{component_name}": component,
+        "%{dctk_comp_generic_headers}": repr(
+            [] if component in _COMPONENTS_DECLARING_OWN_HEADERS else [None],
+        ),
     }
     return _expand_template(repository_ctx, tpl_label, substitutions = substitutions)
 
--- a/cuda/private/templates/BUILD.cccl
+++ b/cuda/private/templates/BUILD.cccl
@@ -55,3 +55,27 @@ cc_library(
     ),
     deps = [":cub"],
 )
+
+# The version-aware `cccl_headers`, replacing the generic one from BUILD.dctk_comp, which
+# always points at `include/`. CUDA 13 moved the CCCL headers into `include/cccl/`, so the
+# generic path resolves `<cccl/thrust/...>` rather than `<thrust/...>`.
+#
+# Deliverable repos only: there, the generic target is suppressed for this component (see
+# template_helper.bzl) and this one takes its place. A local toolkit expands this fragment
+# into @cuda itself, where `%{component_name}` is "cuda" and BUILD.lctk_cuda already owns
+# `cuda_headers`.
+[
+    cc_library(
+        name = "%{component_name}_headers",
+        hdrs = glob(
+            ["%{component_name}/include/**"],
+            allow_empty = True,
+        ),
+        includes = if_cuda_toolkit_version_ge(
+            (13, 0),
+            ["%{component_name}/include/cccl"],
+            ["%{component_name}/include"],
+        ),
+    )
+    for _ in if_local_cuda_toolkit([], [None])
+]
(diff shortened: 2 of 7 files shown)


cloudhan:
I will delay this after #501

